PR #14, opened 2014-08-07
Password Encryption for wildfly-maven-plugin

netdevfighter:
added default mvn decryption functionlly to the plugin
@see http://maven.apache.org/guides/mini/guide-encryption.html


jamezp:
This was added in https://github.com/wildfly/wildfly-maven-plugin/commit/276a2feca4276725640ed1b67869a578ca5eb9f7

PR #23, opened 2014-12-11
Skip parameter for execute-commands goal

dzolo:
Contains support for skip parameter in execute-commands goal that enables to skip commands execution.
This parameter is available in other goals but it is missing in execute-commands goal.


jamezp:
Sorry I'm way behind on these.

dzolo:
Sure I agree.

PR #40, opened 2015-05-13
Introduce ignoreFailures parameter to execute-commands goal

jmartisk:
This is useful for making the goal idempotent (eg. not failing if called repeatedly)

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/cli/Commands.java
+++ b/plugin/src/main/java/org/wildfly/plugin/cli/Commands.java
@@ -81,6 +81,12 @@ public class Commands {
     private List<File> scripts = new ArrayList<File>();
 
     /**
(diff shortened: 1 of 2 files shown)


jamezp:
CLI already has [if-else control](https://developer.jboss.org/wiki/If-elseControlFlow) and [try-catch-finally control](https://developer.jboss.org/wiki/Try-catch-finallyControlFlow). I'd need a compelling reason to add an option to the plugin.

I'm being a bit stricter about it because this plugin has a lot of goals and a lot of settings at this point. At some point I'd like to trim the fat from it and I'm being very cautious about adding new goals and configuration settings.

jmartisk:
I see, didn't know about that. In that case I can use the if-else control. Thanks.

PR #46, opened 2015-05-27
add management user during wildfly:run

rzymek:
Support management user setup during `wildfly:run`.
Configuration:

    <plugin>
        <groupId>org.wildfly.plugins</groupId>
        <artifactId>wildfly-maven-plugin</artifactId>
        <configuration>
            <add-user>
                <user>admin</user>
                <password>admin</password>
            </add-user>
        </configuration>
    </plugin>

This is an alternative to calling 

    bash target/wildfly-run/wildfly-8.2.0.Final/bin/add-user.sh -u admin -p admin

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/common/PropertyNames.java
+++ b/plugin/src/main/java/org/wildfly/plugin/common/PropertyNames.java
@@ -85,4 +85,5 @@ public interface PropertyNames {
 
     String SERVER_ARGS = "wildfly.serverArgs";
 
+    String ADD_USER = "wildfly.addUser";
 }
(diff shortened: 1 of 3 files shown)


jamezp:
I talked to the security team and `org.jboss.sasl.util.UsernamePasswordHashUtil` is likely going away so I'm a bit hesitant to use it in the plugin. However I added the ability to add a user in https://github.com/wildfly/wildfly-maven-plugin/pull/64. The one negative is it does create a new process for each user added, but it's a short lived process and probably something that will not be used all that often.

Please feel free to give feedback on it. I'll try an cut a release today with it so it can be tested.

PR #47, opened 2015-05-30
Allow configuration whether command/script execution should continue in case of an error or not.

sventorben:
Currently maven builds break when one of the commands or scripts executed by the wildfly-maven-plugin fails. However, this is not always desirable. Hence, this pull request introduces a new configuration parameter 'failOnError'. The parameter allows to configure whether execution should continue in case a script/command fails or not.

For example, this can be extremely helpful, when adding/removing modules to wildfly, since it is currently not possible to list or view installed modules via CLI (cf. [WFCORE-430](https://issues.jboss.org/browse/WFCORE-430) or [WFCORE-661](https://issues.jboss.org/browse/WFCORE-661)).
Adding or removing a module via ```module add/remove --name <moduleName> ...``` may fail and break the build (e.g. in automated CI/deployment scenarios) depending on whether the module is present on the AS or not. The configuration parameter 'failOnError' can now be used to continue the build instead of breaking the build in such scenarios.


jamezp:
I'd rather keep that type of logic to CLI with the [if-else-control](https://developer.jboss.org/wiki/If-elseControlFlow) and [try-catch-finally](https://developer.jboss.org/wiki/Try-catch-finallyControlFlow).

sventorben:
The try-catch-finally did not work for me. When adding a module that does already exist (or removing a module that does not exists, respectively) I faced the same issues as described in [this thread](https://developer.jboss.org/message/872851).
If-then-else does not work either, because it is not possible to list or view installed modules via CLI.
That is the reason why I have implemented this 'failOnError' parameter and simply use the [safeHandle](https://github.com/wildfly/wildfly-core/blob/master/cli/src/main/java/org/jboss/as/cli/CommandContext.java#L331) method which is provided by the API.

Helverian:
In our new company project we have a big complex maven script. This feature would help us to handle a problem like "database already created".

shillner:
I'd also think that this feature is really useful in some use-cases. We also have some setups where it would help to continue the build process if a single command failed.

jamezp:
Sorry for the very long delays here. I haven't forgotten about any of the PR's :) I'm hoping to get to processing them in the next couple of weeks and cutting a new release.

sventorben:
Sounds great! If there's anything I can support with, drop me a line.

jamezp:
I'm closing this since it won't cleanly apply anymore. Please see #77 for the new PR though. Please have a look at the new PR and let me know your thoughts. @sventorben I did steal your test so let me know if you'd like me to remove it :)

sventorben:
Feel free to keep the test.
I'll have a look at #77 later this day.

Issue #52, opened 2015-06-26
Support for runtime-name/commonName


voidswitch:
In domain-mode sometimes the same application needs to be deployed in different versions on multiple server-groups. If deployed by hand this can be achieved by setting a different name, but common runtime-name. 

The wildfly DeploymentPlanBuilder supports setting name and commonName, which can be used in DomainDeployment. I've implemented parts of it. As I haven't any knowledge of Maven plugin development, I don't know what more has to be done to make this work. 

see voidswitch/wildfly-maven-plugin@bf37175d4ad401798030fb00cd70968a80cb247f

sventorben:
A collegue of mine just came along with the same issue. Unfortunately, I overlooked your post here and implemented something nearly identical.
Have a look at sventorben/wildfly-maven-plugin@d81ef9395c99f4fd7271f4d68fa37b73a28a599f

The-Alchemist:
Any chance of these being merged?: )

jamezp:
See https://issues.jboss.org/browse/WFMP-16. I'll have a new PR up in a bit since the previous one needs some rebasing. Just testing it now.

PR #54, opened 2015-08-06
Allow to specify runtime name of deployment

sventorben:
This implements wildfly/wildfly-maven-plugin#52

Not sure whether more test or anything else is needed.


shillner:
Vote for this one ;)
I totally need this feature for our development setup. Since we are going to switch production setup to use runtime names, we will also need runtime names for our development setup.

jamezp:
I did the same thing in a new PR https://github.com/wildfly/wildfly-maven-plugin/pull/66. This one needed a rebase so I just redid it. Thanks for the idea and the PR though!

Issue #73, opened 2015-11-11
Allow classifiers for deploy-artifact and undeploy-artifact


shillner:
The goals deploy-artifact and undeploy-artifact are not aware of artifact classifiers. Imagine the following constellation:

**Project A:**
* packaging: war
* attaches the classes as a jar with classifier classes

**Project B:**
* has a compile dependency to a-classes.jar
* needs to deploy the artifact a.war in the integration test phase

**Problem:**
Since the compile dependency to the classes jar has been specified first, the plugin will deploy the classes jar instead of the war which breaks the integration tests.

I will send a pull request for this fix.

jamezp:
I've moved this issue to JIRA https://issues.jboss.org/browse/WFMP-20

PR #74, opened 2015-11-11
added a classifier parameter as well as a check for the classifier fo…

shillner:
…r the deploy-artifact mojos.

For issue: https://github.com/wildfly/wildfly-maven-plugin/issues/73


jamezp:
Thanks for this PR! There were quite a bit of format changes in it though so I made up a new one with some tests https://github.com/wildfly/wildfly-maven-plugin/pull/76.

PR #77, opened 2015-11-25
[WFMP-17] Allow commands to continue being executed if a failure on a previous command has occurred

jamezp:
Since a refactor of the way commands are executed PR #47 will no longer cleanly apply. I've updated the concept here to include the new "local" way to execute commands as well.

One key difference is in this PR batch commands are not safely executed nor are script files.

The main reason batch commands aren't safely handled is it seems odd. If a command in a batch fails it seems like everything should be rolled back.

For script files I think the logic for safely handling commands should likely just be in the script itself. Using a try-catch-finally or if-else should work in script files.

Any feedback on this is welcome. I'd love to hear if this will or will not work for the users that would like this feature.

*P.S.* I'm really sorry for the long delay on this. Once I get some feedback I'll cut a release though.

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/cli/CommandExecutor.java
+++ b/plugin/src/main/java/org/wildfly/plugin/cli/CommandExecutor.java
@@ -107,7 +107,7 @@ public class CommandExecutor {
                 if (commands.isBatch()) {
                     executeBatch(ctx, commands.getCommands());
                 } else {
-                    executeCommands(ctx, commands.getCommands());
+                    executeCommands(ctx, commands.getCommands(), commands.isFailOnError());
                 }
                 executeScripts(ctx, commands.getScripts());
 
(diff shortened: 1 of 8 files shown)


sventorben:
Though I agree regarding batch commands, I think this PR is not feasible when executing scripts.
My concern is basically the same as I have mentioned in PR #47.

[My initial motivation for implementing the PR](https://github.com/wildfly/wildfly-maven-plugin/pull/47#issue-82753421) was that the [try-catch-finally and if-else statements do not work when I try to add modules](https://github.com/wildfly/wildfly-maven-plugin/pull/47#issuecomment-110076832). Since my team is generating scripts for different environments, this PR does not solve our problem.

jamezp:
Excellent feedback, thanks!

With the `execute-commands` goal it will actually use the [`CommandContext.safeHandle()`](https://bluejeans.com/thjenkin/) if the `jbossHome` is not set.

I'm wondering though if local execution, e.g. the `jbossHome` is set, if writing a script was a bad idea. I'll play with piping something to the input for the process which might be a better solution to allow those to be safely handled.

jamezp:
I spent longer than I should have looking at this and I think at this point it's best to pull out the local modular execution stuff. We'll lose the ability to execute commands loaded by extensions, like messaging, but we'll gain better control of safely handling commands.

Probably something more like https://github.com/wildfly/wildfly-maven-plugin/compare/master...jamezp:WFMP-17-b.

I could go either way on scripts too. If you've got a strong opinion either way let me know :)

jamezp:
Replaced by #79.

PR #79, opened 2015-12-03
[WFMP-17] Allow commands to continue being executed if a failure on a…

jamezp:
… previous command has occurred.

This removes the ability to execute commands located in WildFly extensions. See PR #59 and https://issues.jboss.org/browse/WFMP-11 for details on what was implemented. The main reason for the removal is the need to safely handle execution of commands if the user doesn't want the commands the build to break if the commands fail. This seems to be more important at this point than allowing commands in extensions to be executed.

This is also a replacement for #77.


jamezp:
@sventorben does this look better? This will work with normal commands, but not batch runs or scripts. I could change it to work with scripts too, but it seems odd to do if the script contains the `batch` command.

PR #82, opened 2015-12-23
Add --debug Parameter to RunMojo

cainj13:
I found a need for the `--debug` parameter while using this plugin, and after fiddling around with some scripts and what not, just decided to add it.  A couple of notes:

* I chose to separate the core functionality from the test cases (2 commits) since the test cases require a system dependency on the tools.jar.  I wasn't sure if this was a dependency you wanted to introduce, as it will require all contributors to have the JDK installed.  Seems reasonable to me, but I segregated them just in case.  Also, if you know of a way to check whether a server is listening for debuggers that does not use the JDI, it might knock out that dependency.
* It looks like you use several of the same configuration settings from `StartMojo` on `RunMojo`.  I'd be happy to make corresponding changes to `RunMojo` to keep the two in sync.  Just let me know and I'll send another PR your way!

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/common/PropertyNames.java
+++ b/plugin/src/main/java/org/wildfly/plugin/common/PropertyNames.java
@@ -93,4 +93,9 @@ public interface PropertyNames {
 
     String SERVER_ARGS = "wildfly.serverArgs";
 
+    String DEBUG = "wildfly.debug";
+
+    String DEBUG_PORT = "wildfly.debugPort";
+
+    String DEBUG_SUSPEND = "wildfly.debugSuspend";
 }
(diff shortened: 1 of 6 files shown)


cainj13:
This is getting pretty stale... can we make a decision on it one way or the other please?

jamezp:
Yes sorry. I need to get *much* better at reviewing open PR's.

I'm not a huge fan of this. Then again I'm not a huge fan that the `--debug` option is allowed in the scripts :) Even more so here though as with the `start` goal either a standalone or domain instance can be launched.

I also feel in some ways these goals already have too much configuration and adding more seems messy.

It's not quite as elegant from the user point of view, but you could create a debug maven profile with the follow to allow for an easy debug.
```xml
<configuration>
    <java-opts>
        <java-opt>-Xdebug</java-opt>
        <java-opt>-Xnoagent</java-opt>
        <java-opt>-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y</java-opt>
    </java-opts>
</configuration>
```

jamezp:
Well I let this sit a lot longer than I wanted to. I'm sorry for that.

I would also like to thank you for the contribution. I think however I'd rather not add an option at this point since it can already be configured using the `java-opts` configuration.

PR #83, opened 2016-01-06
Prevent NPE if classifier of dependency is null

bakito:
If the classifier of a project dependency is not defined it is null (maven 3.3.3).

This fix prevents a NPE during the verification of the mojo.


ppalaga:
> If the classifier of a project dependency is not defined it is null (maven 3.3.3).

I wonder if the classifier can ever be an empty string as you check?

Similarly, is the trimming really useful?

Otherwise `Objects.equals(a.getClassifier(), classifier)` would do the job.

bakito:
I implemented it just to be save.

If you think Objects.equals is enough, i can change it.

jamezp:
I think I'd prefer `Objects.equals()` though it's not a huge deal.

bakito:
All replaced.

PR #94, opened 2016-04-14
Report server process death during startup properly

ppalaga:


ppalaga:
@jamezp could you please review and eventually merge? This is the very same issue as I fixed in https://github.com/wildfly/wildfly-arquillian/pull/73

jamezp:
I've add similar changes in https://github.com/wildfly/wildfly-maven-plugin/pull/98. This should hopefully give some better error messages.

PR #106, opened 2016-06-14
SAX2 driver class org.apache.xerces.parsers.SAXParser not found

Tibor17:
@jamezp 
Please have a look.
When you plan beta version or final release?

### DIFF
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -175,6 +175,13 @@
             <artifactId>wildfly-protocol</artifactId>
         </dependency>
 
+        <!-- https://github.com/wildfly/wildfly-maven-plugin/issues/105 -->
+        <dependency>
+            <groupId>com.sun.org.apache</groupId>
+            <artifactId>jaxp-ri</artifactId>
+            <version>1.4</version>
+        </dependency>
+
         <!-- Test dependencies -->
         <dependency>
             <groupId>junit</groupId>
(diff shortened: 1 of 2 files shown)


jamezp:
I'll need to look into how all this JAXP stuff works again. I haven't looked at it for a bit. I don't love adding a dependency. It's probably not a massive issue, but it's possible IBM J9 and/or Azul's VM's don't use the RI for JAXP.

As for a release, I'm not sure yet. I've got another major refactor getting ready to go in so hopefully I'll get another alpha out this week. If it seems stable I'll do a beta and a release soon after.

Tibor17:
@jamezp 
I agree this dependency and SAX driver fetching is just ugly patch.
I see `ctx.toBatchedCommand(cmd)` needs to have only String as command, so then why we use JAXP if we can use JAXB which is part of standard Java API.
Can I create JAXB stubs and write three lines of parser for your plugin?
WDYT?

Tibor17:
@jamezp 
Can CommandContext be configured for SAX driver?

jamezp:
It more than likely uses the StAX. Changing it would be a pretty big change and likely rejected. We'd also need a new release of WildFly Core 1.x which is unlikely to happen. It's best to try to fix this in the plugin itself if it's possible.

jamezp:
I've looked into this in detail now. The details of my findings are on [JIRA comment](https://issues.jboss.org/browse/WFMP-48?focusedCommentId=13262852&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13262852).

I really don't like bringing in a default JAXP implementation, but it's definitely a solution. I think I'd prefer using a forked processes if I can figure out how to reliably read the output.

I'm going to close this for now as it would need a rebase anyway. I'm going to explore forking the process again.

Another option is to not initialize the modular environment if the `<jboss-home/>` is not set. While it doesn't make this issue go away, it does limit it's scope.

PR #107, opened 2016-06-15
[WFMP-41] Create a new core project and refactor plugin to use the co…

jamezp:
…re project. Cleaned up the utilities for the API.


core/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+    <scm>
+        <connection>${wildfly.scm.connection}</connection>
+        <developerConnection>${wildfly.scm.developer.connection}</developerConnection>
+        <url>${wildfly.scm.url}</url>
+    </scm>
+
+    <build>
+        <plugins>
+        </plugins>
+    </build>
```
ppalaga: Redundant block, that can perhaps be removed.

core/src/main/java/org/wildfly/plugin/core/ContainerDescription.java
```diff
@@ -0,0 +1,71 @@
...
+     * @return the release version or {@code null} if not defined
+     */
+    String getReleaseVersion();
+
+    /**
+     * Returns the type of the server that was launched.
+     *
+     * @return the type of the server that was launched or {@code null} if not defined
+     */
+    String getLaunchType();
```
ppalaga: It would be interesting either to name the possible launch type values or put a link here to where the values are listed.

core/src/main/java/org/wildfly/plugin/core/DefaultContainerDescription.java
```diff
@@ -0,0 +1,139 @@
...
+        final ModelNode result = client.execute(op);
+        if (Operations.isSuccessfulOutcome(result)) {
+            final ModelNode model = Operations.readResult(result);
+            final String productName = getValue(model, "product-name", "WildFly");
+            final String productVersion = getValue(model, "product-version");
+            final String releaseVersion = getValue(model, "release-version");
+            final String launchType = getValue(model, "launch-type");
+            return new DefaultContainerDescription(productName, productVersion, releaseVersion, launchType, "DOMAIN".equalsIgnoreCase(launchType));
+        } else {
+            Logger.getLogger(DefaultContainerDescription.class).errorf("Failed to read the root resource: ", Operations.getFailureDescription(result));
```
ppalaga: I guess that `UNKNOWN` will be useless in most situations. Why not throwing an exception instead of returning `UNKNOWN`?

core/src/main/java/org/wildfly/plugin/core/Deployment.java
```diff
@@ -0,0 +1,299 @@
...
+
+    /**
+     * Creates a new deployment for the file. If the file is a directory the content will be deployed exploded using
+     * the file system location.
+     *
+     * @param content the file containing the content
+     *
+     * @return the deployment
+     */
+    public static Deployment of(final File content) {
```
ppalaga: Do we need both all these `Deployment.of(...)` overloads AND the corresponding public constructors? There may exist reasons to hide the constructors behind `of(...)`, but why keeping both (or even keeping both as public)?

core/src/main/java/org/wildfly/plugin/core/Deployment.java
```diff
@@ -0,0 +1,299 @@
...
+ */
+@SuppressWarnings({"unused", "WeakerAccess"})
+public class Deployment implements DeploymentDescription {
+
+    private final Path pathContent;
+    private final InputStream inputStreamContent;
+    private final Set<String> serverGroups;
+    private String name;
+    private String runtimeName;
+    private boolean enabled;
```
ppalaga: Why are `name`, `runtimeName` and `enabled` not final?

core/src/main/java/org/wildfly/plugin/core/Deployment.java
```diff
@@ -0,0 +1,299 @@
...
+
+    /**
+     * Adds a server group for the deployment.
+     *
+     * @param serverGroup the server group to add
+     *
+     * @return this deployment
+     */
+    public Deployment addServerGroup(final String serverGroup) {
+        serverGroups.add(serverGroup);
```
ppalaga: I vote for making Deployment immutable. It is a model class that can be shared among multiple threads. `serverGroups.add(serverGroup)` should go to `DeploymentBuilder` and `serverGroups` should be passed unmodifiable from there: `new Deployment(..., Collections.unmodifiableSet(serverGroups))`.

core/src/main/java/org/wildfly/plugin/core/Deployment.java
```diff
@@ -0,0 +1,299 @@
...
+     * @return this deployment
+     */
+    public Deployment addServerGroups(final Collection<String> serverGroups) {
+        this.serverGroups.addAll(serverGroups);
+        return this;
+    }
+
+    @Override
+    public Set<String> getServerGroups() {
+        return new LinkedHashSet<>(serverGroups);
```
ppalaga: Yes, returning a copy makes serverGroups unmodifiable, but not thread safe.

core/src/main/java/org/wildfly/plugin/core/DeploymentException.java
```diff
@@ -0,0 +1,81 @@
...
+        super(message);
+    }
+
+    /**
+     * Creates a new deployment exception.
+     *
+     * @param format the {@linkplain String#format(String, Object...) format} for the message
+     * @param args   the arguments for the format
+     */
+    public DeploymentException(final String format, final Object... args) {
```
ppalaga: Having both `DeploymentException(final String format, final Object... args)` and `public DeploymentException(final String message, final Throwable cause)` may lead to ambiguities in usage. I vote for removing  `DeploymentException(final String format, final Object... args)` or alternatively, we could change the  order of the args in `DeploymentException(final String message, final Throwable cause)` to DeploymentException(final Throwable cause, final String message)`.

core/src/main/java/org/wildfly/plugin/core/DeploymentManager.java
```diff
@@ -0,0 +1,203 @@
...
+     * @param serverGroup the server group to check for the deployment on
+     *
+     * @return {@code true} if the deployment exists otherwise {@code false}
+     */
+    boolean hasDeployment(String name, String serverGroup) throws IOException;
+
+    /**
+     * A factory to create a new deployment manager
+     */
+    class Factory {
```
ppalaga: BTW, seeing factory here, have you considered using dependency injection? As you may know, JSR 330 is available in maven. I am not sure what is the situation like in your other target environments, but at least Gradle seems to have it too.

core/src/main/java/org/wildfly/plugin/core/DeploymentOperations.java
```diff
@@ -0,0 +1,479 @@
...
+     * Note this does not deploy the content.
+     * </p>
+     *
+     * @param content     the path to the content to deploy
+     * @param name        the name for the deployment content, if {@code null} the paths file name is used
+     * @param runtimeName the runtime name for the deployment content
+     *
+     * @return the add operation
+     */
+    public static Operation createDeploymentAddOperation(final Path content, final String name, final String runtimeName) {
```
ppalaga: Do we need this one, given that we have `createDeployOperation(final Deployment deployment)` ?

core/src/main/java/org/wildfly/plugin/core/DeploymentOperations.java
```diff
@@ -0,0 +1,479 @@
...
+     * Note this does not deploy the content.
+     * </p>
+     *
+     * @param content     the content to deploy
+     * @param name        the name for the deployment content
+     * @param runtimeName the runtime name for the deployment content, can be {@code null}
+     *
+     * @return the add operation
+     */
+    public static Operation createDeploymentAddOperation(final InputStream content, final String name, final String runtimeName) {
```
ppalaga: Do we need this one, given that we have `createDeployOperation(final Deployment deployment)` ?

core/src/main/java/org/wildfly/plugin/core/Deployment.java
```diff
@@ -0,0 +1,299 @@
...
+     * @return the path to the deployment content
+     */
+    Path getPathContent() {
+        return pathContent;
+    }
+
+    /**
+     * The input stream representing the deployment content.
+     * <p>
+     * Note that if {@link #usePathContent()} returns {@code true} this will return {@code null}.
```
ppalaga: I think that having to check `usePathContent()` makes the API unnecessarily complicated. For the sake of cleanness and avoiding potential mistakes when using this API, I think it would be worth thinking of a way to have just one field to carry the "source" info. (The bottom line is that when handing over io streams among classes, one must take care to define clearly, who is responsible for closing them, to avoid leaks.) 

I see two options:

(a) Use `java.net.URL` that is both easily mappable from either Path/File and most InputStreams have an associated URL. I admit that the *most* in the previous sentence might not be enough and that because `java.net.URL` is final, and that `URLStreamHandler` s are complicated to use, the option (b) might be better.

(b) Introduce our own `DeploymentSource` interface with a single `InputStream openStream()` method.

With `DeploymentSource` it would be natural to expect that who calls `DeploymentSource.openStream()` is also responsible for closing it.

ppalaga:
First review round finished.

core/src/main/java/org/wildfly/plugin/core/Deployment.java
jmazzitelli: I would reject (a) here. In my use case, I have an InputStream that does not map to a URL - it is coming directly to me as a InputStream - I do not have a URL.

core/src/main/java/org/wildfly/plugin/core/ContainerDescription.java
jamezp: It may not be consistent depending on the target container. For example WildFly 8 doesn't have `SELF_CONTAINED`. This API isn't meant to be a full management API. Just a helper for some small things.

core/src/main/java/org/wildfly/plugin/core/DefaultContainerDescription.java
jamezp: Good point. I copied this from elsewhere and didn't want the error there, but here it does make sense.

core/src/main/java/org/wildfly/plugin/core/Deployment.java
jamezp: I mainly just had both to satisfy users. Some people prefer `new` some prefer static factories. I'll think about limiting the scope though.

core/src/main/java/org/wildfly/plugin/core/Deployment.java
jamezp: I don't see this being used in a multi-threaded environment often. The only situation where I could see it being useful is if you're deploying the same content to several servers at once. I will document that this is not thread-safe however.

Also if it becomes immutable it would require either a lot of constructors or a builder. I'm not real keen on making yet another builder for an API that's really supposed to be pretty simple and here for convenience.

core/src/main/java/org/wildfly/plugin/core/Deployment.java
jamezp: This is all for internal usage only. It's not ideal, but it's a simple solution for now. The only reason I don't use just an `InputStream`, which could be done for a file quite easily, is the `OperationBuilder`'s allow for a file to be attached. This can be useful in some cases as IIRC the entire stream doesn't always need to be copied.

I'll also document that the streams are *always* closed by this API.

core/src/main/java/org/wildfly/plugin/core/DeploymentException.java
jamezp: Ah, good point. I usually just add these for convenience, but they're not really necessary.

core/src/main/java/org/wildfly/plugin/core/DeploymentManager.java
jamezp: I'm not sure that would work as the `DeploymentManager` needs a `ModelControllerClient` which would also need a producer. I can't think of a good way to create a producer for the `ModelControllerClient`.

core/src/main/java/org/wildfly/plugin/core/DeploymentOperations.java
jamezp: It's here for convenience only. It's useful in cases where you just want to add content, but not deploy it. This may be more common in a managed domain where you just want to add content to the repository, but not add it to any server groups.

core/src/main/java/org/wildfly/plugin/core/Deployment.java
jamezp: Well the more I think about it something like a `DeloymentStream` might be needed. If the original `InputStream` is closed the `Deployment` instance will no longer be usable for any deploy operations. At the least, which I might just do for now, I'll need to wrap in the `InputStream` in one that will allow reuse.

core/src/main/java/org/wildfly/plugin/core/ContainerDescription.java
ppalaga: OK, then, to make the JavaDoc a bit useful, we can add just that: `The possible values returned by this method depend on the underlying container - e.g. for WildFly 10, the values are SELF_CONTAINED, ... whatever else`

core/src/main/java/org/wildfly/plugin/core/Deployment.java
ppalaga: > I don't see this being used in a multi-threaded environment often.

You never know what crazy piping and threading a build tool can come with. And this is supposed to be used by build tools primarily, is it not? I think this is doable easily enough so that we can afford it :)

core/src/main/java/org/wildfly/plugin/core/DeploymentManager.java
ppalaga: You may well be right in this particular situation (that I have not looked at in detail). The question about dependency injection just came to my mind, when I saw the Factory - because Factories are generally often eliminated by using DI. I must say I am not sure myself, if DI is a good idea for a lib like this that is supposed to be used in Maven, Gradle, SBT and who knows where else, where a DI container is not always available. 

And finally, to make the picture complete, why I am bringing this topic: Since longer, I planned the very same refactoring in my srcdeps-maven-plugin[https://github.com/l2x6/srcdeps-maven-plugin] . Much like you, I am introducing a srcdeps-core lib so that it can be used not only from Maven but also from Gradle and other tools. 
I started two days ago, and I have chosen to use the DI. I am curious to figure out myself if that turns out to be a good idea :) Here is the working branch if you happen to want to have a look: https://github.com/ppalaga/srcdeps-maven-plugin/tree/160613-introduce-core

jmazzitelli: FYI: I'll be using this in the hawkular agent, and it doesn't use DI.

ppalaga: > I'll be using this in the hawkular agent

Agent Mazz cannot be ignored ;)

core/src/main/java/org/wildfly/plugin/core/DeploymentOperations.java
ppalaga: >  add content, but not deploy it

Ah, I did not know this scenario exists.

core/src/main/java/org/wildfly/plugin/core/Deployment.java
jamezp: After thinking about this it would be too complicated to make it thread safe. Even if immutable the underlying stream is mutable. If two threads try to deploy the same content, assuming an `InputStream` was used, it could deploy invalid content. One thread could be blocking on a read while the other thread reads a chunk, then the waiting thread could get a lock and read data from the wrong position.

PR #113, opened 2016-07-15
Add support for properties files, like the CLI client does.

Jurrie:
In the process, I removed some spurious throws clauses and converted some return types from Collection to List (because a List is ordered).

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/cli/CommandExecutor.java
+++ b/plugin/src/main/java/org/wildfly/plugin/cli/CommandExecutor.java
@@ -23,11 +23,15 @@
 package org.wildfly.plugin.cli;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


jamezp:

Jurrie:
Yes, I agree to the MIT License terms.

jamezp:
I think this look good. One concern I do have is since this sets system properties, other plugins/goals that are executed after this will also see the system properties. I may add some code to remove the system properties set after the goal is executed as part of another PR.

And just to reference the JIRA https://issues.jboss.org/browse/WFMP-51.

jamezp:
Just an FYI I've refactor this a bit in https://github.com/wildfly/wildfly-maven-plugin/pull/120. I moved the configuration to be outside the `<execute-commands/>` tag. I'm actually going to [deprecate](https://issues.jboss.org/browse/WFMP-54) the `<execute-commands/>` configuration property.

Basically instead of
```xml
<configuration>
    <execute-commands>
        <propertiesFiles>
            <propertiesFile>some.properties</propertiesFile>
        </propertiesFiles>
    </execute-commands>
</configuration>
```

You'd just do
```xml
<configuration>
    <propertiesFiles>
        <propertiesFile>some.properties</propertiesFile>
    </propertiesFiles>
</configuration>
```

PR #116, opened 2016-07-27
[WFMP-36] Deprecate complex Domain configuration property.

jamezp:


jamezp:
Note that the tests needed to set the new configuration parameters because the maven test harness doesn't do well when a parameter has a property associated with it. The test pom's still include the configuration values, but they're ignored for now.

PR #139, opened 2017-02-24
Use new process when executing offline CLI

ctomc:


ctomc:
@jamezp updated the PR to check for result of the script execution.
and if it fails it display up to 3 lines of execution log.
Better would be to have CLI outout error to std err, but as long as we don't have that this should work just fine.

PR #140, opened 2017-03-01
Added option to redirect the output of offline CLI command execution.…

jamezp:
… Also ensure commands run as well as scripts.

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/cli/ExecuteCommandsMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/cli/ExecuteCommandsMojo.java
@@ -27,6 +27,7 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


jamezp:
@ctomc sorry for the delay on this, would you mind giving it a review to see if it will work for you? If so I'll merge it and release it first thing.

You can get the same result you had by using `<stdout>${project.basedir}/target/wildfly-plugin/script-out.log</stdout>` or simply just use `<stdout>none</stdout>` and nothing will be printed unless there was an error.

plugin/src/main/java/org/wildfly/plugin/common/StandardOutputStream.java
```diff
@@ -0,0 +1,188 @@
...
+                target = Target.DISCARDING;
+                out = ServerProcess.DISCARDING;
+            } else {
+                target = Target.COLLECTING;
+                out = new ByteArrayOutputStream();
+            }
+        } else {
+            target = Target.FILE;
+            // Attempt to create a file
+            stdoutPath = Paths.get(value);
```
ctomc: Only thing I am not 100% about is this. It would make sense to put file under "target",
OTOH it does make sense to have "full path" here.

plugin/src/main/java/org/wildfly/plugin/cli/ExecuteCommandsMojo.java
```diff
@@ -121,24 +122,46 @@
...
+     * <ul>
+     * <li>{@code none} indicates the {@code stdout} and {@code stderr} stream should not be consumed</li>
+     * <li>{@code System.out} or {@code System.err} to redirect to the current processes <em>(use this option if you
+     * see odd behavior from maven with the default value)</em></li>
+     * <li>Any other value is assumed to be the path to a file and the {@code stdout} and {@code stderr} will be
+     * written there</li>
+     * </ul>
+     * </div>
+     */
+    @Parameter(name = "stdout", defaultValue = "System.out", property = PropertyNames.STDOUT)
```
ctomc: Maybe default would be better as target/wildfly-plugin.log

jamezp: I think for now I'm just going to leave it as `System.out` because it's the same default behavior for non-offline output. However making the default something like `${project.basedir}/target/wildfly-plugin/script-out.log` may make sense.

PR #143, opened 2017-06-06
add support for java opts when executing offline CLI

ctomc:


plugin/src/main/java/org/wildfly/plugin/cli/ExecuteCommandsMojo.java
```diff
@@ -177,7 +180,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         if (offline) {
             getLog().debug("Executing offline CLI scripts");
             try (StandardOutputStream out = StandardOutputStream.parse(stdout, false)) {
-                final int exitCode = offlineCLIExecutor.execute(jbossHome, getCommands(), getLog(), out, systemProperties);
+
+                String[] opts = javaOpts.split("[\\n\\s]+");
```
ctomc: This here is so we can support properties passes in multiple lines as well as just separated by space same as surefire does it

jamezp: `\s` already matches `\n`.

jamezp: From the JavaDoc:
> \s	A whitespace character: [ \t\n\x0B\f\r]

ctomc: Hmm, not sure why it gave me issues when I was testing it...

jamezp: Do you have an example string by chance? I'm just curious is all.

ctomc: https://github.com/wildfly/wildfly/blob/master/pom.xml#L6291

jamezp: Thanks. With a simple test I see the same results with or without the `\n`. I'll give the actually build a test and see what happens though. Other than that I think this PR is fine and I'll get it merged here soon.

PR #145, opened 2017-06-20
[WFMP-76] Rework how console streams are consumed. Preferring to use …

jamezp:
…a ProcessBuilder.Redirect rather than consuming the streams.

https://issues.jboss.org/browse/WFMP-76


jamezp:
Note the removal of the {{ServerProcess}} from the core module. This seemed like a good idea, but would require changes and releases every time the `java.lang.Process` API changes. This would make this "helper" not very helpful.

PR #150, opened 2017-08-02
WFMP-84 fix NPE when using path-elements for modulesPath

markusdy:
See [WFMP-84](https://issues.jboss.org/browse/WFMP-84) for further explanation.


jamezp:
This looks reasonable.

markusdy:
Of course, I do.

PR #157, opened 2017-09-11
[WFMP-88, WFCORE-3272] stdout redirect to file creates directory stru…

xstefank:
…cture in lower case

WFMP: https://issues.jboss.org/browse/WFMP-88
WFCORE: https://issues.jboss.org/browse/WFCORE-3272

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/common/StandardOutput.java
+++ b/plugin/src/main/java/org/wildfly/plugin/common/StandardOutput.java
@@ -125,7 +125,7 @@ public class StandardOutput {
             }
         } else {
             // Attempt to create a file
-            stdoutPath = Paths.get(value);
+            stdoutPath = Paths.get(stdout.trim());
             if (Files.notExists(stdoutPath)) {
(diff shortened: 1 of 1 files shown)


plugin/src/test/java/org/wildfly/plugin/common/StandardOutputTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class StandardOutputTest {
+
+    @Test
+    public void testStdoutFileContainsUpperCaseLetters() throws IOException {
```
jamezp: This test fails on Windows as paths aren't case sensitive. I'm fine with either just removing the test or using an Assume to ignore the test on Windows.

xstefank: It's not necessary just used as part of TDD, thank you for the clarification

PR #168, opened 2018-05-17
WildFly 12 upgrade

flashboss:
WildFly 10 is dated. Better work with the last new stable version


jamezp:
That version is only used for testing. I've left it at WildFly 10 mainly to ensure/test backwards compatibility.

jamezp:
I'm closing this for now. The 10.1.0.Final version is targeted for backwards compatibility.

PR #186, opened 2019-01-24
Extended deployment with option for unmanaged archives

karsie:


jamezp:
I think this seems to make sense. I'm away this week, but hopefully next week I'll have a better chance to review it.

PR #189, opened 2019-02-14
Add include-packagings property to whitelist package types.

leluna:
Sometimes, deploying all jars is not desired. Also, there may be additional packaing types such as maven-archetype which should not be processed. Therefore, I have created a whitelist property to define which packaging types should be included. It is defined as a comma separated String property to make it usable from CLI.

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/common/PropertyNames.java
+++ b/plugin/src/main/java/org/wildfly/plugin/common/PropertyNames.java
@@ -57,6 +57,8 @@ public interface PropertyNames {
 
     String ID = "wildfly.id";
 
+    String INCLUDE_PACKAGINGS = "wildfly.includePackagings";
+
     String IGNORE_MISSING_DEPLOYMENT = "undeploy.ignoreMissingDeployment";
 
     String JAVA_HOME = "java.home";
--- a/plugin/src/main/java/org/wildfly/plugin/deployment/AbstractAppDeployment.java
+++ b/plugin/src/main/java/org/wildfly/plugin/deployment/AbstractAppDeployment.java
@@ -23,6 +23,8 @@
 package org.wildfly.plugin.deployment;
 
 import java.io.File;
+import java.util.Arrays;
+import java.util.List;
 
 import org.apache.maven.plugins.annotations.Parameter;
 import org.wildfly.plugin.common.PropertyNames;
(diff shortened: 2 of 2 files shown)


jamezp:
Generally speaking I think it's best to use the `<skip>true</true>` in a parent pom and use the `<skip>false</skip>` in the pom of the module you want deployed. You could also control it with a property like `<skip>${skip.deploy}</skip>` which defaults to `true` and you'd just add the property to the poms you want where you override to `false`.

leluna:
We are not using declaring the plugin in our POMs at all, the plugin is executed in Jenkins as a post build step directly. We are deploying our sample applications for testing, and do not want to use the plugin directly.

jamezp:
Do you by chance have an example of how that works? I'm not all that familiar with Jenkins and just want to understand the problem a little better.

jamezp:
I think I have another idea. What do you think of using the project name instead? Something like:
```
<includes>
    <include>my-ejb</include>
    <include>my-war</include>
</includes>
```

Then just match the project name to those. If it's not in the list then it's ignored.

jamezp:
I'd need to add a JIRA and create some tests, but this is what I was thinking https://github.com/wildfly/wildfly-maven-plugin/compare/master...jamezp:temp.

leluna:
We directly execute the goal from CLI after a successful maven build, e.g. 

> mvn -U org.wildfly.plugins:wildfly-maven-plugin:2.1.0.Beta1-SNAPSHOT:deploy-only

I was thinking about using the <includes> pattern first, but unfortunately it is not possible to pass a list property from CLI, or is it? If it is possible, that would be much better than the comma separated list.

Whitelisting projects would also be okay for us, though it would be less convenient. Automatically including all wars gives us the convenience that the deployments are "auto discovered".

jamezp:
Ah okay. Thanks for the conversation BTW :) It definitely helps the full picture.

So maybe we could use a GAV or something with a pattern. Like:
```
<includes>
    <include>${project.groupId}:*:*:war</include>
</includes>
```

jamezp:
From the command line I suppose it could just be
```
mvn -U org.wildfly.plugins:wildfly-maven-plugin:2.1.0.Beta1-SNAPSHOT:deploy-only -Dwildfly.deployment.includes=*:*:*:war,*.*ejb:jar
```

Seems like that should work :) Probably what I should have used instead of the pattern for undeploy.

leluna:
That looks cool! (Just discovered my search for "arrays as command line arguments for maven plugin" was not exhausive enough, or at least reading all the answers from StackOverflow whould have helped *facepalm*.)

jamezp:
Okay I created https://issues.jboss.org/browse/WFMP-115. I really appreciate the idea and input here. I'm going to close this for now until we make a decision on the best way to implement this.

leluna:
Thanks for the quick response!

PR #200, opened 2021-10-01
[WFMP-130] Deprecate parameters that refer to downloading a Maven artifact for the server. These will be removed in 3.0.0.

jamezp:
https://issues.redhat.com/browse/WFMP-130

Also https://issues.redhat.com/browse/WFMP-132

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,7 @@
         <version.org.wildfly.checkstyle-config>1.0.7.Final</version.org.wildfly.checkstyle-config>
         <version.org.wildfly.common.wildfly-common>1.5.1.Final</version.org.wildfly.common.wildfly-common>
         <version.org.wildfly.core>9.0.1.Final</version.org.wildfly.core>
-        <version.org.wildfly.dist>14.0.1.Final</version.org.wildfly.dist>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -83,7 +83,7 @@
         <version.org.wildfly.checkstyle-config>1.0.7.Final</version.org.wildfly.checkstyle-config>
         <version.org.wildfly.common.wildfly-common>1.5.1.Final</version.org.wildfly.common.wildfly-common>
         <version.org.wildfly.core>9.0.1.Final</version.org.wildfly.core>
-        <version.org.wildfly.dist>14.0.1.Final</version.org.wildfly.dist>
+        <version.org.wildfly.dist>24.0.1.Final</version.org.wildfly.dist>
```
jmesnil, line 86: @jamezp  I'm not sure why you need to bump WildFly version but should the core version be updated too?

If that's the case, WildFly 24.0.1.Final incorporated WildFly Core 16.0.1.Final.

jamezp: Upgraded to 17.0.1.Final https://github.com/wildfly/wildfly-maven-plugin/pull/201. The main reason for the upgrade from 14 is testing with Java 17 so I went with the latest release :)

PR #212, opened 2022-01-11
Fix for WFMP-142, Plugin should handle the case of not found deployment

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-142

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
@@ -318,7 +318,7 @@ public class PackageServerMojo extends AbstractProvisionServerMojo {
         }
     }
 
-    private Path getDeploymentContent() {
+    private Path getDeploymentContent() throws MojoExecutionException {
         final PackageType packageType = PackageType.resolve(project);
         final String filename;
         if (this.filename == null) {
@@ -326,7 +326,18 @@ public class PackageServerMojo extends AbstractProvisionServerMojo {
         } else {
             filename = this.filename;
         }
-        return targetDir.toPath().resolve(filename);
+        Path deployment = targetDir.toPath().resolve(filename);
+        if (Files.notExists(deployment)) {
+            if (this.filename != null ) {
+                throw new MojoExecutionException("No deployment found wih name " + this.filename);
+            }
+            if (this.runtimeName != null ) {
+                throw new MojoExecutionException("No deployment found although a runtime-name has been set. "
+                        + "Check the final name of the deployment. A custom file name can be set with <filename> parameter.");
+            }
+            getLog().warn("No deployment file found, the server will not contain a deployment");
+        }
+        return deployment;
     }
 
     private static void cleanupServer(Path jbossHome) throws IOException {
--- a/tests/standalone-tests/src/test/java/org/wildfly/plugin/provision/PackageTest.java
+++ b/tests/standalone-tests/src/test/java/org/wildfly/plugin/provision/PackageTest.java
@@ -25,6 +25,7 @@ package org.wildfly.plugin.provision;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import org.apache.maven.plugin.Mojo;
+import org.apache.maven.plugin.MojoExecutionException;
 import org.junit.Test;
 import org.junit.Assert;
 import org.wildfly.plugin.tests.AbstractProvisionConfiguredMojoTestCase;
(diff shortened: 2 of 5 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -318,15 +318,26 @@ private void warnExtraConfig(Path extraContentDir) {
...
         final PackageType packageType = PackageType.resolve(project);
         final String filename;
         if (this.filename == null) {
             filename = String.format("%s.%s", project.getBuild().getFinalName(), packageType.getFileExtension());
         } else {
             filename = this.filename;
         }
-        return targetDir.toPath().resolve(filename);
+        Path deployment = targetDir.toPath().resolve(filename);
+        if (!Files.exists(deployment)) {
```
jamezp, line 330: Very minor, but you could uses `Files.notExists(deployment)`. It just makes it slightly more readable.

jamezp (approved):

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
jfdenise: Thanks James, fixed.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -318,15 +318,26 @@ private void warnExtraConfig(Path extraContentDir) {
...
+        Path deployment = targetDir.toPath().resolve(filename);
+        if (Files.notExists(deployment)) {
+            if (this.filename != null ) {
+                throw new MojoExecutionException("No deployment found wih name " + this.filename);
+            }
+            if (this.runtimeName != null ) {
+                throw new MojoExecutionException("No deployment found although a runtime-name has been set. "
+                        + "Check the final name of the deployment. A custom file name can be set with <filename> parameter.");
+            }
+            getLog().warn("No deployment file found, the server will not contain a deployment");
```
mchoma, line 338: Can plugin log, what was he looking for? That mean to log `filename`.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -318,15 +318,26 @@ private void warnExtraConfig(Path extraContentDir) {
...
         }
-        return targetDir.toPath().resolve(filename);
+        Path deployment = targetDir.toPath().resolve(filename);
+        if (Files.notExists(deployment)) {
+            if (this.filename != null ) {
+                throw new MojoExecutionException("No deployment found wih name " + this.filename);
+            }
+            if (this.runtimeName != null ) {
+                throw new MojoExecutionException("No deployment found although a runtime-name has been set. "
+                        + "Check the final name of the deployment. A custom file name can be set with <filename> parameter.");
```
mchoma, line 336: Can plugin log, what was he looking for? That mean to log filename.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -318,15 +318,26 @@ private void warnExtraConfig(Path extraContentDir) {
...
             filename = this.filename;
         }
-        return targetDir.toPath().resolve(filename);
+        Path deployment = targetDir.toPath().resolve(filename);
+        if (Files.notExists(deployment)) {
+            if (this.filename != null ) {
+                throw new MojoExecutionException("No deployment found wih name " + this.filename);
+            }
+            if (this.runtimeName != null ) {
+                throw new MojoExecutionException("No deployment found although a runtime-name has been set. "
```
mchoma, line 335: At first sight It was confusing to me why `runtime-name` is mentioned. I assume it is because when it is set, that mean user want match some deployment. So my understanding of error message is something like this

```
"No deployment " + filename +"found. As runtime-name has been set, that indicate that there is expected some deployment. A custom path to deployment can be set with <filename> parameter."
```

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
jfdenise: @mchoma in this case the filename is null. It can be that no deployment has been set (imagine a pom packaging). This is just a log to advertise that no deployment has been found and deployed.

jfdenise: But I see your point, we could still log what we tried to retrieve.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
jfdenise: I will make small adjustement to the fix. Thank-you for your feedback.

PR #214, opened 2022-01-13
Fix for WFMP-144, Can package goal decide values for inherit-packages…

jfdenise:
…, inherit-configs itself

### DIFF
--- a/core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
+++ b/core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
@@ -120,8 +120,25 @@ public class GalleonUtils {
             List<String> excludedLayers,
             Map<String, String> pluginOptions) throws ProvisioningException, IllegalArgumentException {
         final ProvisioningConfig.Builder state = ProvisioningConfig.builder();
-        for (FeaturePack fp : featurePacks) {
+        boolean hasLayers = !layers.isEmpty();
+        boolean fpWithDefaults = true;
+        if (!hasLayers) {
+            // Check we have all feature-packs with default values only.
+            for (FeaturePack fp : featurePacks) {
+                if (fp.isInheritConfigs() != null ||
+                        fp.isInheritPackages() != null ||
+                        !fp.getIncludedConfigs().isEmpty() ||
+                        !fp.getExcludedConfigs().isEmpty() ||
+                        fp.isTransitive() ||
+                        !fp.getExcludedPackages().isEmpty() ||
+                        !fp.getIncludedPackages().isEmpty()) {
+                    fpWithDefaults = false;
+                    break;
+                }
+            }
+        }
 
+        for (FeaturePack fp : featurePacks) {
             if (fp.getLocation() == null && (fp.getGroupId() == null || fp.getArtifactId() == null)
                     && fp.getNormalizedPath() == null) {
                 throw new IllegalArgumentException("Feature-pack location, Maven GAV or feature pack path is missing");
@@ -140,12 +157,25 @@ public class GalleonUtils {
             final FeaturePackConfig.Builder fpConfig = fp.isTransitive() ? FeaturePackConfig.transitiveBuilder(fpl)
                     : FeaturePackConfig.builder(fpl);
             if (fp.isInheritConfigs() == null) {
-                fpConfig.setInheritConfigs(false);
+                if (hasLayers) {
+                    fpConfig.setInheritConfigs(false);
+                } else {
+                    if (fpWithDefaults) {
+                        fpConfig.setInheritConfigs(true);
+                    }
+                }
             } else {
                 fpConfig.setInheritConfigs(fp.isInheritConfigs());
             }
+
             if (fp.isInheritPackages() == null) {
-                fpConfig.setInheritPackages(false);
+                if (hasLayers) {
+                    fpConfig.setInheritPackages(false);
+                } else {
+                    if (fpWithDefaults) {
+                        fpConfig.setInheritConfigs(true);
+                    }
+                }
             } else {
                 fpConfig.setInheritPackages(fp.isInheritPackages());
             }
--- a/plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
@@ -230,8 +230,10 @@ abstract class AbstractProvisionServerMojo extends AbstractMojo {
             }
             getLog().info("Provisioning server in " + home);
             pm.provision(config);
-            if (!Files.exists(home)) {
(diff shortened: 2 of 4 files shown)


jfdenise:
@mchoma are you fine with the wording of the message when configuring the feature-packs for default config?

core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
```diff
@@ -111,17 +111,43 @@ public static ProvisioningConfig buildConfig(Path provisioningFile) throws Provi
...
+                        !fp.getExcludedConfigs().isEmpty() ||
+                        fp.isTransitive() ||
+                        !fp.getExcludedPackages().isEmpty() ||
+                        !fp.getIncludedPackages().isEmpty()) {
+                    fpWithDefaults = false;
+                    break;
+                }
+            }
+            if (fpWithDefaults) {
+                log.log(LogConsumer.Type.INFO, "No Layers have been provided and feature-packs don't override default "
```
mchoma, line 141: I understand what you mean. But for more fluent reading I would suggest something like:

```
No Layers are configured and feature-packs don't override default packages and configs inheritance. That mean the last feature-pack default configuration will be provisioned.

In case that is not what you would expect, you can configure inheritance of configurations in feature-packs using <inherit-configs>true|false</inherit-configs>
```

core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
```diff
@@ -111,17 +111,43 @@ public static ProvisioningConfig buildConfig(Path provisioningFile) throws Provi
...
+                        + "packages and configs inheritance. An attempt is done to provision the last feature-pack default configuration. "
+                        + "In case that is not what you would expect, make sure to properly inherit configurations "
+                        + "from the right feature-pack using <inherit-configs>true|false</inherit-configs>");
+            }
+        }
 
+        for (int i = 0; i < featurePacks.size(); i++) {
+            FeaturePack fp = featurePacks.get(i);
+            boolean isLastFP = (i == featurePacks.size() - 1);
```
mchoma, line 150: So last is really last as mentioned in list of FPs. So when I put to list also datasource feature pack, this default mechanism will be broken?

Is it dumped somewhere in plugin run what `inherit-*` were in fact used?

jfdenise: Yes, don't put the datasources as the last one. But the datasources shouldn't be used, it has just layers, no default configs.

core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
```diff
@@ -140,10 +166,21 @@ public static ProvisioningConfig buildConfig(ProvisioningManager pm,
...
+                            fpConfig.setInheritConfigs(true);
+                        } else {
+                            fpConfig.setInheritConfigs(false);
+                        }
+                    }
+                }
             } else {
                 fpConfig.setInheritConfigs(fp.isInheritConfigs());
             }
+            // We are optimistic, all default configs should be defined with Layers so we don't need to inherit default packages.
```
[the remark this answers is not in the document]
jfdenise: Yes, that is correct. The default configs defined fully with layers (that is the goal), don't require to inherit packages. When inheriting packages, all JBoss modules get installed even the not needed ones.

core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
jfdenise: Thank-you, I adjusted the wording.

core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
mchoma: Just for me to understand. So `inherit-config=true` cant be set on datasources? Or can but will be ignored or do nothing.

So logic here when layers.isEmpty() then mark `inherit-config=true` on last FP will ignore datasource fp?

core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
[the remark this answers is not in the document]
mchoma: Just for me to understand. There is some mapping between config-->packages?

If layer has defined config snippet, that automatically determine which packages it brings in?

That is why one can use config when talking about packages, because config identify packages unambiguously?

core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
jfdenise: Yes, it could be set, but nothing will get provisioned, there are no default configs defined in this feature-pack. 
We can't ignore an FP, we have no knowledge of the FP. If the datasources FP is the latest, nothing will get proviioned with the default logic.

mchoma: Ok got it. So therefore this would be important to troubleshoot similar cases

`Is it dumped somewhere in plugin run output what values of inherit-* were in fact used?`

jfdenise: Yes, the information can be found in the server. if record-provisioning-state is set to true, then galleon generates a .galleon/provisioning.xml file.
If record-provisioning-state is not set (the default), we generate a .wildfly-maven-plugin-provisioning.xml file that contains this information.

jamezp (approved):
I don't want to block merging this as it makes sense. However, I don't love the `LogConsumer` part, yet I also don't have a better idea. I think we can merge this though and I'll have a think about how it could be handled.

jfdenise:
@jamezp , thanks for the review. I need to rework this PR, actually we will not need this automatic detection, so no need for the logger ;-)
I will rework it early next week.

jfdenise:
I updated the PR to enable inheritance in all cases if no layers have been provided and the feature-packs have not been customized.

core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
```diff
@@ -140,12 +157,25 @@ public static ProvisioningConfig buildConfig(ProvisioningManager pm,
...
                     : FeaturePackConfig.builder(fpl);
             if (fp.isInheritConfigs() == null) {
-                fpConfig.setInheritConfigs(false);
+                if (hasLayers) {
+                    fpConfig.setInheritConfigs(false);
+                } else {
+                    if (fpWithDefaults) {
+                        fpConfig.setInheritConfigs(true);
+                    }
+                }
```
jamezp, line 166: I assume the default here is `false`. I'm just wondering if this should be `fpConfig.setInheritConfigs(!hasLayers && fpWithDefaults)`.

jamezp (approved):

PR #215, opened 2022-01-28
Fix for WFMP-145, Introduce an offline CLI command executor

jfdenise:
* Introduce a BaseCommandConfiguration CLI configuration class and builder.
* Refactor CommandConfiguration to inherit from BaseCommandConfiguration and expose builder.
* Introduce an AbstractCommandExecutor.
* Refactor the CommandExecutor to inherit from the AbstractCommandExecutor
* Introduce an OfflineCommandExecutor to inherit from the AbstractCommandExecutor.
* OfflineCommandExecutor is injected into the package goal
* CommandExecutor is injected into other goals (unchanged).


jamezp:
I thought there was a plan to keep a dependency on the WildFly Launcher. Is that not the case?

jamezp (approved):

PR #222, opened 2022-03-17
Fix for WFMP-148, Support for iterative development workflow

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-148

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
@@ -22,7 +22,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
```diff
@@ -265,4 +282,28 @@ static Path resolvePath(MavenProject project, Path path) {
         return path;
     }
 
+    public void setFeaturePacksFromProperty() {
+        String fps = System.getProperty(PropertyNames.WILDFLY_PROVISIONING_FEATURE_PACKS);
```
jamezp, line 286: How would this be used? I don't think it's wrong or anything, I'm just trying to understand the use-case.

jfdenise: @jamezp , it would be used in this kind of context: https://github.com/wildfly/wildfly-proposals/pull/461/files#diff-81512f32e1f2215e55b0a246a066be2a621704c1d3932621c5963d312b3a8c0dR208

jamezp: You could add a `FeaturePack.set(String)` method which would allow it to be set via a property. I suppose in some ways this is the same, but it's just not the standard.

jamezp: You'd just add something like:
```java
    public void set(final String value) {
        setLocation(value);
    }
```

IIRC Maven will handle the comma split for you.

jamezp: That makes sense. Do you want this merged for now and to it later or would you rather implement that now?

jfdenise: @jamezp , I will implement it now.

jfdenise: @jamezp , the feature-packs are now set in the standard maven way.

jamezp (approved):

PR #226, opened 2022-04-08
Next version is 3.0.1.Final-SNAPSHOT

jmesnil:


jmesnil:
@jamezp I'm not sure if that was intentional to drop the `Final` from the version.

PR #236, opened 2022-06-15
Fix for WFWIP-460, When extra-content-dir is a regular file, build pr…

jamezp:
…ocess fails without explanation

https://issues.redhat.com/browse/WFMP-162


jmesnil:
Upstream PR: https://github.com/wildfly/wildfly-maven-plugin/pull/227

PR #238, opened 2022-06-15
[WFMP-161] Allow the feature pack location to be overridden for the s…

jamezp:
…tart and run goals.

https://issues.redhat.com/browse/WFMP-161

Upstream #237

### DIFF
--- a/core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
+++ b/core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
@@ -21,6 +21,8 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+
 import org.jboss.galleon.Constants;
 import org.jboss.galleon.ProvisioningDescriptionException;
 import org.jboss.galleon.ProvisioningException;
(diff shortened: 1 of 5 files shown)


plugin/src/main/java/org/wildfly/plugin/server/RunMojo.java
```diff
@@ -360,7 +371,7 @@ private Path provisionIfRequired(final Path installDir) throws MojoFailureExcept
         try {
             if (!Files.exists(installDir)) {
                 getLog().info("Provisioning default WildFly server in " + installDir);
```
jfdenise, line 373: Perhaps we could change this sentence to: "Provisioning default server in"

core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
```diff
@@ -247,9 +248,9 @@ private static String getMavenCoords(FeaturePack fp) {
         return builder.toString();
     }
 
-    private static String getWildFlyFeaturePackLocation(String version) {
+    private static String getWildFlyFeaturePackLocation(String featurePackLocation, String version) {
         StringBuilder fplBuilder = new StringBuilder();
-        fplBuilder.append(WILDFLY_DEFAULT_FEATURE_PACK_LOCATION);
+        fplBuilder.append(featurePackLocation);
```
jfdenise, line 253: We could throw an exception if featurepackLocation is null.

plugin/src/main/java/org/wildfly/plugin/server/RunMojo.java
jamezp: +1 that makes sense.

core/src/main/java/org/wildfly/plugin/core/GalleonUtils.java
jamezp: Good point. When I tested manually it did error out, but being more explicit probably makes more sense. I wonder if we should validate the pattern too.

PR #253, opened 2022-09-09
[WFMP-174] New image goal to build (and push) application image

jmesnil:
As an user, I want to build and push an application image from the plugin

https://issues.redhat.com/browse/WFMP-174

- [x] new `image` goal
- [x] new tests in standalone-tests
- [x] code is commented (and generated in the web site)
- [x] code is documented (with an example in the generated web site)

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
@@ -1,0 +1,221 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.wildfly.plugin.provision;
+
+import static java.lang.String.format;
+import static java.lang.String.join;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.Duration;
+import java.util.Arrays;
+
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+
+/**
+ * Build (and push) an application image containing the provisioned server and the deployment.
+ * <p>
+ * The {@code image} goal extends the {@code package} goal, building and pushing the image occurs after the server
+ * is provisioned and the deployment deployed in it.
+ * <p>
+ * The {@code image} goal relies on a Docker binary to execute all image commands (build, login, push).
+ *
+ * @since 4.0
+ */
+@Mojo(name = "image", requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME, defaultPhase = LifecyclePhase.PACKAGE)
+public class ApplicationImageMojo extends PackageServerMojo {
+
+    public static final int DOCKER_CMD_CHECK_TIMEOUT = 3000;
+
+    /**
+     * The configuration of the application image.
+     * <p>
+     * The {@code image} goal accepts the following configuration:
+     * <p>
+     * <pre>
+     * &lt;image&gt;
+     *   &lt;!-- (optional) set it to false to skip build the application image (true by default) --&gt;
+     *   &lt;build&gt;true&lt;/build&gt;
+     *
+     *   &lt;!-- (optional) set it to true to (login and) push the application image to the container registry (false by default).
+     *
+     *     If user and password are not specified, the image goal will not attempt to login to the container
+     *     registry prior to pushing the image.
+     *     The login to the container registry must then be performed before Maven is run.
+     *     --&gt;
+     *   &lt;push&gt;true&lt;/push&gt;
+     *
+     *   &lt;!-- (optional) The binary used to perform image commands (build, login, push) (default is "docker") --&gt;
+     *   &lt;docker-binary&gt;docker&lt;/docker-binary&gt;
+     *
+     *   &lt;!-- (optional) the JDK version used by the application. Default is "11", Allowed values are "11" and "17 --&gt;
+     *   &lt;jdk-version&gt;11&lt;/jdk-version&gt;
+     *
+     *   &lt;!-- (optional) The group part of the name of the application image --&gt;
+     *   &lt;group&gt;${user.name}&lt;/group&gt;
+     *
+     *   &lt;!-- (optional) The name part of the application image. If not set, the value of the artifactId (in lower case) is used --&gt;
+     *   &lt;name&gt;${project.artifactId}&lt;/name&gt;
+     *
+     *   &lt;!-- (optional) The tag part of the application image (default is "latest") --&gt;
+     *   &lt;tag&gt;latest&lt;/tag&gt;
+     *
+     *   &lt;!-- (optional) The container registry. If set, the registry is added to the application name.
+     *     If the image is pushed and the registry is not set, it defaults to "docker.io" to login to the registry
+     *     --&gt;
+     *   &lt;registry&gt;quay.io&lt;/registry&gt;
+     *
+     *   &lt;!-- (optional) The user name to login to the container registry (if push is enabled). --&gt;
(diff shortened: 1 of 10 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+
+@Mojo(name = "image", requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME, defaultPhase = LifecyclePhase.PACKAGE)
+public class ApplicationImageMojo extends PackageServerMojo {
+
+    public static final int DOCKER_CMD_CHECK_TIMEOUT = 3000;
+    private static final String DOCKER_BINARY = "docker";
```
ehsavoie, line 45: Why not make this a parameter so that we can use podman if we want/need to ?

jmesnil: As mentioned, this is a nice-to-have that can be considered. I can't run `podman` on my machine to test it right now.

Podman suggest to alias it to `docker` so this can be a workaround in the mean time.
If you test that it works with the alias, I could more confidently make the binary configurable too.

ehsavoie: I can do that :)

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * The configuraiton of the application image.
```
ehsavoie, line 48: typo

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+
+@Mojo(name = "image", requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME, defaultPhase = LifecyclePhase.PACKAGE)
+public class ApplicationImageMojo extends PackageServerMojo {
+
+    public static final int DOCKER_CMD_CHECK_TIMEOUT = 3000;
```
iweiss, line 44: Could this be a system property or a plugin config option?

jmesnil: This timeout *only* applies to run `docker -v` (to test that `docker` is present). That's plenty of time and I don't see a need to make it configurable. wdyt?

iweiss: I think that's fair for local installations. Do you think CI/CD runs or podman via Docker API compatibility (local socket) could be affected?

jmesnil: Even on CI, if `docker -v` takes 3 seconds to execute you might have bigger issues to fix first : )

iweiss: Very true. Just erring on the side of caution.

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
jmesnil: done, you can configure it with `<dockerBinary>` (default is `docker`)

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
jmesnil: thanks

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
```diff
@@ -0,0 +1,220 @@
...
+ * Build (and push) an application image containing the provisioned server and the deployment.
+ *
+ * The {@code}image{@code} goal extends the {@code}package{@code} goal, building and pushing the image occurs after the server
+ * is provisioned and the deployment deployed in it.
+ *
+ * The {@code}image{@code} goal relies on a Docker binary to execute all image commands (build, login, push).
+ *
+ * @since 4.0
+ */
+@Mojo(name = "image", requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME, defaultPhase = LifecyclePhase.PACKAGE)
```
jfdenise, line 52: I am wandering if we could have an option to disable actual build of the image but still generate the Dockerfile. A kind of dry-run, allowing to build docker image outside of maven or use a maven plugin.

jmesnil: That's a good idea. The Dockerfile is generated in the `target`. I could generate it in any case when the `image` goal is run and check for `image.build` after its generation.

jmesnil: done, @jfdenise thanks for the suggestion.

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
```diff
@@ -0,0 +1,223 @@
...
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+
+/**
+ * Build (and push) an application image containing the provisioned server and the deployment.
+ *
```
jamezp, line 44: Just a minor thing. These empty lines will be removed from the documentation and JavaDocs. It's best to use do something like:
```
//**
 * Build (and push) an application image containing the provisioned server and the deployment.
 * <p>
 * The {@code image} goal extends the {@code package} goal, building and pushing the image occurs after the server
 * is provisioned and the deployment deployed in it.
 * </p>
 * <p>
 * The {@code image} goal relies on a Docker binary to execute all image commands (build, login, push).
 * </p>
 * @since 4.0
 */
 ```

File: plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java [hunk 6]
```diff
@@ -0,0 +1,223 @@
...
+ * The {@code}image{@code} goal extends the {@code}package{@code} goal, building and pushing the image occurs after the server
```
jamezp, line 45: These `{@code}value{@code}` should be changed to `{@code value}` or `<code>value</code>`.

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
```diff
@@ -0,0 +1,223 @@
...
+                boolean pushSuccess = pushApplicationImage(image);
+                if (!pushSuccess) {
+                    throw new MojoExecutionException(String.format("Unable to push application image %s", image));
+                }
+                getLog().info(String.format("Successfully pushed application image %s", image));
+            }
+        } catch (IOException e) {
+            MojoExecutionException ex = new MojoExecutionException(e.getLocalizedMessage());
+            ex.initCause(e);
+            throw ex;
```
jamezp, line 157: Couldn't we just use the `new MojoExecutionException(e.getLocalizedMessage(), e);`?

plugin/src/main/java/org/wildfly/plugin/provision/ExecUtil.java
```diff
@@ -0,0 +1,161 @@
...
+     * @param directory The directory
+     * @param command The command
+     * @param args The command arguments
+     * @return true if commands where executed successfully
+     */
+    public static boolean exec(Log log, File directory, String command,
+                               String... args) {
+        try {
+            Process process = startProcess(directory, command, args);
+            new HandleOutput(process.getInputStream(), log).run();
```
jamezp, line 81: Is there a reason we're not launching this in a new thread? It seems like we should to ensure we get all the content.

jamezp (commented):
Just some minor comments mostly related to documentation.

One thing that could be interesting to add too, in a later PR even, is to if the password is empty prompt the user for one. We do something like this in `ClientCallbackHandler` using the `System.console()`.

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
jmesnil: fixed, thanks

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
jmesnil: fixed, thanks

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
jmesnil: as you can see, I don't write Javadoc quite often :D

plugin/src/main/java/org/wildfly/plugin/provision/ExecUtil.java
jamezp: I definitely could be wrong, but won't it block until the process is complete? I guess this one is done without a timeout so maybe not a big deal and the next `process.waitFor()` would block anyway. It probably doesn't matter, just looked strange to me I suppose :)

jmesnil: yes, this one will block until the process is complete.
There is no timeout on this one (eg builder the docker image can take as many time as it needs)

PR #256, opened 2022-10-06
[WFMP-175] Add property to skip the deployment in the package goal.

jmesnil:
If wildfly.package.deployment.skip is true, the user deployment will not be deployed in the provisioned server.

JIRA: https://issues.redhat.com/browse/WFMP-175

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
@@ -160,6 +160,12 @@ public class PackageServerMojo extends AbstractProvisionServerMojo {
     @Parameter(defaultValue = "false", property = PropertyNames.SKIP_PACKAGE)
     private boolean skip;
 
+    /**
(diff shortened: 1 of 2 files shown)


jfdenise (approved):

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -160,6 +160,13 @@ public class PackageServerMojo extends AbstractProvisionServerMojo {
     @Parameter(defaultValue = "false", property = PropertyNames.SKIP_PACKAGE)
     private boolean skip;
 
+    /**
+     * Set to {@code true} if you want to skip deploying the deployung in the provisioned server.
```
jamezp, line 164: Is "deployung" supposed to be "deployment"?

jmesnil: typing too fast! I've updated the javadoc, thanks

jamezp (approved):

PR #272, opened 2022-11-30
Push image from project build directory path instead of hard target path

staerion:
Why the pull request?

We're currently using the wildfly plugin from a child pom. When we call the child pom from the parent pom, the plugin wants to push the image from the directory <parent_pom>/target, which does not exist. Only the directory <child_pom>/target exists. Generating and building the docker file all happens from the project build directory as well, so hopefully this makes sense.


jmesnil (approved):

jmesnil:
@staerion thanks for the bug fix!

PR #281, opened 2023-01-25
WFMP-177 and WFMP-178

jfdenise:
Issues: 
* https://issues.redhat.com/browse/WFMP-177
* https://issues.redhat.com/browse/WFMP-178

* Latest changes introduced in wildfly-channel 1.0.0.Beta4
* Use prospero project to generate metadata file when channels are setup.

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
@@ -48,7 +48,6 @@ import org.jboss.galleon.universe.maven.repo.MavenRepoManager;
 import org.jboss.galleon.util.IoUtils;
 import org.jboss.galleon.xml.ProvisioningXmlWriter;
 import org.wildfly.channel.UnresolvedMavenArtifactException;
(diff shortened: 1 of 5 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
```diff
@@ -178,7 +177,7 @@ abstract class AbstractProvisionServerMojo extends AbstractMojo {
     String layersConfigurationFileName;
 
     @Parameter(alias = "channels", required = false)
-    List<ChannelCoordinate> channels;
+    List<ChannelConfiguration> channels;
```
jamezp, line 180: What are the differences here? Are there parameter name differences? If so we'll need to change the version to something like 4.1.0 before this can be merged.

jfdenise: @jamezp , yes we have an incompatible change there, we introduce a sub element `manifest-coordinate` inside a Channel. +1 for 4.1.0, I was going to suggest it.

jamezp (approved):

PR #286, opened 2023-02-07
** [WFARQ-???] Check if podman exists and use that first, otherwise u…

jamezp:
…se docker.

Signed-off-by: James R. Perkins <jperkins@redhat.com>


vsalbaba:
It would be even better if the actual executable name could be provided as a parameter on the command line

jamezp:
> It would be even better if the actual executable name could be provided as a parameter on the command line, however, I consider this to be better than the current state.

I think we could do something like `System.getProperty("wildfly.image.dockerBinary")`. I'm not sure that would honor anything defined in the `<properties/>` section of a POM. The standard properties do not work for complex configuration types in maven.

vsalbaba:
There should also be updates to the documentation, in many places it implies that docker is the binary that will be used

https://github.com/wildfly/wildfly-maven-plugin/search?q=docker

jmesnil:
@vsalbaba what would be the name to use to categorise both `docker` and `podman`?

jamezp:
We can't change the `<dockerBinary/>` without breaking compatibility. I do see some documentation that should be updated, but I don't think we should change the configuration parameters.

vsalbaba:
There is also a hardcoded mention at https://github.com/wildfly/wildfly-maven-plugin/blob/e11d2096957b06171e0c28fab69785cdbbea3373/tests/standalone-tests/src/test/java/org/wildfly/plugin/provision/ImageTest.java#L49

jamezp:
> There is also a hardcoded mention at
> 
> https://github.com/wildfly/wildfly-maven-plugin/blob/e11d2096957b06171e0c28fab69785cdbbea3373/tests/standalone-tests/src/test/java/org/wildfly/plugin/provision/ImageTest.java#L45-L50

Thank you. Just fixed that.

vsalbaba:
I am considering that changing the default from `docker` to `podman` might not be a compatible change, as some users may rely on `docker` is the default option. To resolve this issue, we can make `podman` the fallback option and keep `docker` as the default. 

Additionally, I am considering the option of allowing the tests to run with both engines. By using `mvn test -Dwildlfy.image.binary=docke`r, the tests should run with `docker`, even if `podman` is installed. Although the magic guess feature is convenient, I would also like to have more control over which engine is used.

jamezp:
> I am considering that changing the default from `docker` to `podman` might not be a compatible change, as some users may rely on `docker` is the default option. To resolve this issue, we can make `podman` the fallback option and keep `docker` as the default.
> 
> Additionally, I am considering the option of allowing the tests to run with both engines. By using `mvn test -Dwildlfy.image.binary=docke`r, the tests should run with `docker`, even if `podman` is installed. Although the magic guess feature is convenient, I would also like to have more control over which engine is used.

I think this makes sense. I'll update the PR once I get a chance.

plugin/src/site/markdown/image-example.md.vm
```diff
@@ -3,8 +3,9 @@
 The `image` goal allows you to build an application image that contains a server and your application.
 This application image contains your application ready to run on the cloud in a containerized platform such as Kubernetes.
 
-The operations related to the image are executing using a Docker binary. By default, it uses `docker` but it is possible to configure it
-with the `<image><docker-binary>` element to use another binary such as `podman`.
+The operations related to the image are executing using a Docker binary. By default, there is an attempt to determine if
+`podman` is installed. If not, `docker` is used. This can be explicitly set with the `<image><docker-binary>` element.
+It can also be overridden with the `wildfly.image.dockerBinary` property.
```
jmesnil, line 8: the name of the sys prop is now `wildfly.image.binary`

jmesnil (changes requested):
LGTM outside of the name of the sysprop

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageInfo.java
```diff
@@ -98,4 +101,21 @@ String getWildFlyRuntimeImage() {
                 return "quay.io/wildfly/wildfly-runtime-jdk11:latest";
         }
     }
+
+    public String getDockerBinary() {
+        if (dockerBinary == null) {
+            try {
+                if (ExecUtil.execSilentWithTimeout(Duration.ofSeconds(3), "podman", "-v")) {
+                    dockerBinary = "podman";
```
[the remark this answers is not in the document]
jamezp: I do think we should use docker first, then podman to keep the current behavior. I also think it does likely make sense to log the binary being used so the user knows.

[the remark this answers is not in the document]
jamezp: Oh, wait that is already being done :)
```java
getLog().info(format("Building application image %s using %s.", image, this.image.getBinary()));
```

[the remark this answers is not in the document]
jmesnil: @jamezp that's a good point. And `docker` remains the most popular binary to build image.

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageInfo.java
```diff
@@ -77,9 +77,10 @@ public class ApplicationImageInfo {
     protected String password;
 
     /**
-     * The binary used to build and push images (default is "docker").
+     * The binary used to build and push images. The default will attempt to be determined. First we'll check if
+     * {@code podman} is available, if not {@code docker} will be used.
      */
-    protected String dockerBinary = "docker";
+    private String dockerBinary;
```
rhusar, line 83: Let's find better name even for internals, something like `containerTool`, `containerBinary`, `podBinary`, etc. Ideas?

rhusar (changes requested):

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageInfo.java
jamezp: +1, that is what I'm doing. I'm just using `binary` in the `ApplicationImageInfo` as it is typically referred to as "image" so "binary" just seemed to make sense.

rhusar: Gotcha, oh well...

PR #287, opened 2023-02-14
[WFMP-182] Update WildFly runtime image names

jmesnil:
Use the new naming convention for the WildFly runtime image. They all have the same name, `quay.io/wildfly/wildfly-runtime` and uses tags to specify the JDK version.

By default, `jdkVersion` is no longer set so that the WildFLy runtime image will use its latest tag by default. This tag corresponds to the most recent LTS JDK version provided by WildFly (11 at the moment, eventually 17, etc).

User can stay on a specify JDK version by setting <jdk-version> to 11 or 17.

JIRA: https://issues.redhat.com/browse/WFMP-182

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageInfo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageInfo.java
@@ -38,10 +38,11 @@ public class ApplicationImageInfo {
     protected boolean push = false;
 
     /**
-     * Determine which WildFly Runtime image to use so that the application runs with the specified JDK.
-     * The default is "11". Accepted values are "11", "17".
+     * Determine which WildFly runtime image to use so that the application runs with the specified JDK.
+     * If the value is not set, the `latest` tag of the WildFly runtime image is used.
+     * Accepted values are "11", "17".
      */
-    private String jdkVersion = "11";
+    private String jdkVersion;
 
     /**
      * The group part of the name of the application image.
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageInfo.java
```diff
@@ -91,11 +92,16 @@ String getApplicationImageName(String artifactId) {
...
+        final String runtimeImageTag;
         switch (jdkVersion) {
             case "17":
-                return "quay.io/wildfly/wildfly-runtime-jdk17:latest";
+            case "11":
+                runtimeImageTag = "latest-" + jdkVersion;
+                break;
             default:
-                return "quay.io/wildfly/wildfly-runtime-jdk11:latest";
+                runtimeImageTag = "latest";
```
jamezp, line 103: Does this effectively mean the default is 17?

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageInfo.java
```diff
@@ -91,11 +92,16 @@ String getApplicationImageName(String artifactId) {
     }
 
     String getWildFlyRuntimeImage() {
+        String runtimeImageName = "quay.io/wildfly/wildfly-runtime:";
+        final String runtimeImageTag;
         switch (jdkVersion) {
             case "17":
-                return "quay.io/wildfly/wildfly-runtime-jdk17:latest";
+            case "11":
+                runtimeImageTag = "latest-" + jdkVersion;
```
jfdenise, line 100: Should be "latest-jdk"+jdkVersion

jmesnil: fixed, thanks

jamezp (approved):

PR #288, opened 2023-02-16
Fix for WFMP-184, Rework Channel configuration

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-184


jamezp (approved):
I don't know enough about channels, but this seems okay to me.

PR #289, opened 2023-02-20
[WFMP-185] Deploy the application by copying it to deployments dir

jmesnil:
Instead of using the `deploy` management operation, the deployment is copied into the standalone/deployments directory.

The runtime-name field is now used to rename the file that is copied to the standalone/deployments directory.

JIRA: https://issues.redhat.com/browse/WFMP-185


jfdenise (approved):

darranl:
Are we not loosing some functionality from the plugin with this?  Is it not possible for this command to also be used with a remote server?

darranl:
Ah I see this is about packaging the server, it is not the main deploy command

jamezp (approved):

PR #290, opened 2023-02-22
[WFMP-186] Optimize the application image generated by the image goal

jmesnil:
In the image goal, skip the deployment when packaging WildFly

In the generated Dockerfile, use 2 different COPY commands:

* COPY the server bits. This Docker layer can be cached (as long as the server provisioning does not change)
* COPY the deployment in the standalone/deployment directory. This Docker layer is small (the size of the deployment) and will be changed every time

With that optimization, repeatedly pushing the application image will mean to push the deployment Docker layer only.

This optimization saves time, bandwith and storage on the container registry.

JIRA: https://issues.redhat.com/browse/WFMP-186

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
@@ -111,6 +111,10 @@ public class ApplicationImageMojo extends PackageServerMojo {
 
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
+        // when the application image is built, the deployment step is skipped.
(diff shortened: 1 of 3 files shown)


jamezp (approved):

jamezp:
Nothing to do with this PR, but just noticed it. Should we drop 26 support and testing or should we keep it?

jmesnil:
> Nothing to do with this PR, but just noticed it. Should we drop 26 support and testing or should we keep it?

I'd recommend to keep it. WildFly 26 is the last release that supports Jakarta EE 9. We should leave more time to users to update their app to Jakarta EE 10 and WildFly 27+.

jamezp:
If @jfdenise is okay with this, we can merge it.

jfdenise (approved):

PR #291, opened 2023-02-24
[WFMP-191] Add a dev goal for rapid development of WAR files.

jamezp:
https://issues.redhat.com/browse/WFMP-191

@jmesnil @jfdenise What do you guys think of something like this? This isn't yet complete, but it's been something I've wanted to do for a long time. It would differ a bit from the bootable-jar `dev-watch` goal in that we just need to build the WAR and possibly redeploy it.

There are currently some issues with IDE's like IDEA which aggressively save and it keeps trying to compile invalid sources. I'm not completely sure how to handle that. One idea I had was to keep a queue of changed files and only compile when a compile is not already happening. Maybe using a timer of some sort. I've not looked to see how the bootable JAR plugin handles this.

Anyway, just looking for some feedback on the idea.

### DIFF
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -156,6 +156,12 @@
             <groupId>org.jboss.logmanager</groupId>
             <artifactId>jboss-logmanager</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.twdata.maven</groupId>
+            <artifactId>mojo-executor</artifactId>
+            <version>${version.org.twdata.maven}</version>
+        </dependency>
         <dependency>
             <groupId>org.wildfly.core</groupId>
             <artifactId>wildfly-controller-client</artifactId>
--- a/plugin/src/main/java/org/wildfly/plugin/common/Environment.java
+++ b/plugin/src/main/java/org/wildfly/plugin/common/Environment.java
@@ -138,6 +138,15 @@ public class Environment {
         return exe;
     }
 
+    /**
+     * Determines if this is a Windows environment.
+     *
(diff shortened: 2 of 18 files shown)


jfdenise:
@jamezp that is great. Some notes:
* Would you stop/re-provision/re-start the server if one change the plugin configuration (eg add layers) in pom.xml file?
* There are some files to ignore (Hidden files, file that start with a ~).
* I suspect that a lot of logic is similar to WildFly Bootable JAR. But possibly not easy to share.

jamezp:
> @jamezp that is great. Some notes:
> 
> * Would you stop/re-provision/re-start the server if one change the plugin configuration (eg add layers) in pom.xml file?

I was thinking of, at least initially, not supporting POM changes. Only changes to source and resource files would be supported.

> * There are some files to ignore (Hidden files, file that start with a ~).

Okay, good to know. I did notice that. I do have auto-save enabled on IDEA too so maybe that's part of it. It doesn't seem to be a problem, just logs a bunch of compilation error logs.

> * I suspect that a lot of logic is similar to WildFly Bootable JAR. But possibly not easy to share.

I definitely borrowed heavily from the Bootable JAR plugin :) I think there are some common goals like compile and the WAR plugin, but that's about it. I was going to initially only support WAR's too. We could look at supporting JAR's and EAR's later, but I don't know how often that would be used. EAR's seem more difficult to support, but maybe not.

jamezp:
@jfdenise I'm wondering if this should end up being an option in the `run` goal which is enabled by default. Some simple parameter like `watchForChanges`. What do you think?

jfdenise (approved):

PR #306, opened 2023-03-15
Fix for WFMP-195, Fallback behavior when channels are configured is incorrect

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-195

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/ChannelMavenArtifactRepositoryManager.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/ChannelMavenArtifactRepositoryManager.java
@@ -50,14 +50,17 @@ public class ChannelMavenArtifactRepositoryManager implements MavenRepoManager, 
 
     private final ChannelSession channelSession;
     private final List<Channel> channels = new ArrayList<>();
-
+    private final boolean originalVersionResolution;
+    private final Log log;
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/ChannelMavenArtifactRepositoryManager.java
```diff
@@ -84,18 +87,28 @@ public void resolve(MavenArtifact artifact) throws MavenUniverseException {
...
-            // unable to resolve the artifact through the channel.
-            // if the version is defined, let's resolve it directly
-            if (artifact.getVersion() == null) {
+            if (originalVersionResolution) {
+                log.warn("Resolution of artifact " + artifact.getGroupId() + ":" +
+                        artifact.getArtifactId() + " failed. Using original version.");
+                // unable to resolve the artifact through the channel.
+                // if the version is defined, let's resolve it directly
+                if (artifact.getVersion() == null) {
+                    log.warn("No version provided.");
```
jamezp, line 96: Should this actually be a error since it throws an exception?

jfdenise: Good catch James, Fixed!

jamezp (approved):

PR #323, opened 2023-04-11
[WFMP-203] Ensure we use a relative directory when configuring the CO…

jamezp:
…PY command in the Dockerfile.

https://issues.redhat.com/browse/WFMP-203


jmesnil:
thanks James!

PR #341, opened 2023-05-24
Use prospero BOM to manage shared dependencies

spyrkob:
Uses prospero parent as BOM to allign shared dependency versions

Requires https://github.com/wildfly-extras/prospero/pull/401 

cc @jfdenise

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,15 +89,12 @@
 
     <properties>
         <!-- WildFly/JBoss dependencies -->
-        <version.org.jboss.logging.jboss-logging>3.5.0.Final</version.org.jboss.logging.jboss-logging>
-        <version.org.jboss.logmanager>2.1.19.Final</version.org.jboss.logmanager>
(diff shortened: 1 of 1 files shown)


jfdenise:
@jamezp , I see a benefit to align our provisioning tooling (prospero and this plugin provisioning capabilities). We are quickly becoming out of sync.

jamezp:
I would prefer a separate BOM I guess. It seems odd for the Maven plugin to reply on prospero only provisioning uses it. The deploy goals, CLI goals, etc. do not need provisioning. 

Basically prospero is about channels and not Maven. This project is about Maven and not channels. It feels backwards to me.

jmesnil:
I agree with @jamezp. The provisioning capability of the plug-is one among others and I don't think Prospero should be the component that dictates the plugins dependency updates.

spyrkob:
OK I'll work on separate BOM then, thanks for feedback

PR #342, opened 2023-05-26
[WFMP-208]: dev mode deployment runtime name is not configurable.

ehsavoie:
* Adding a new 'name' parameter with the same property as for the package mojo.

Jira: https://issues.redhat.com/browse/WFMP-208

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/dev/DevMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/dev/DevMojo.java
@@ -342,6 +342,14 @@ public class DevMojo extends AbstractServerStartMojo {
     @Parameter(alias = "channels")
     private List<ChannelConfiguration> channels;
 
+    /**
+     * Specifies the name used for the deployment.
+     *
+     * When the deployment is copied to the server, it is renamed with this name.
+     */
+    @Parameter(property = PropertyNames.DEPLOYMENT_NAME)
+    private String name;
+
     // Lazily loaded list of patterns based on the ignorePatterns
     private final List<Pattern> ignoreUpdatePatterns = new ArrayList<>();
     // Lazy loaded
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/wildfly/plugin/dev/DevMojo.java
```diff
@@ -837,7 +845,8 @@ private Deployment getDeploymentContent() {
         if (remote) {
             return Deployment.of(Path.of(project.getBuild().getDirectory()).resolve(filename));
         } else {
-            return Deployment.of(resolveWarDir()).setRuntimeName(filename);
+            String runtimeName = this.name == null || this.name.trim().isBlank() ? filename : this.name;
```
jamezp, line 848: There is no need to use `trim()` with `isBlank()`

plugin/src/main/java/org/wildfly/plugin/dev/DevMojo.java
```diff
@@ -342,6 +342,14 @@ public class DevMojo extends AbstractServerStartMojo {
     @Parameter(alias = "channels")
     private List<ChannelConfiguration> channels;
 
+    /**
+     * Specifies the name used for the deployment.
+     *
+     * When the deployment is copied to the server, it is renamed with this name.
+     */
+    @Parameter(property = PropertyNames.DEPLOYMENT_NAME)
```
jamezp, line 350: I'm on my phone so it's hard to see, but if the run goal has this field, we should remove it from there. Or just move this field into the `DevMojo`.

jamezp (changes requested):

plugin/src/main/java/org/wildfly/plugin/dev/DevMojo.java
ehsavoie: yes it has but not the StartMojo also the RunMojo has runtimeName which is deperecated in the PackageServerMojo

jamezp (approved):

jamezp (changes requested):
This looks like it has formatting issues.

plugin/src/main/java/org/wildfly/plugin/dev/DevMojo.java
ehsavoie: @jamezp I don't see the point of adding this field to the 'parent' class AbstractServerStartMojo and it doesn't make sense in the context of StartMojo. So I don't think we should 'move' it out :)

jamezp:
Replaces by #350 to get this change into the next release.

PR #343, opened 2023-05-30
Fix for WFMP-209, Can't use maven http transport with maven < 3.9

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-209


jamezp:
> @jamezp , the remaining failure is un-related to this change. This build image tests failed in some previous runs.

Yes, I've been thinking about disabling these tests for Windows as they seem flaky. I guess I should dig into why they are flaky first though :)

tests/standalone-tests/pom.xml
```diff
@@ -61,17 +61,17 @@
         <dependency>
             <groupId>org.apache.maven.resolver</groupId>
             <artifactId>maven-resolver-api</artifactId>
-            <scope>test</scope>
+            <scope>provided</scope>
```
jamezp, line 64: What's the reason for these being changed from test to provided? TBH it looks weird to me to have them at all.

jfdenise: They are required for compilation, they were transitively included.

jamezp: There isn't anything that's a non-test that should need to be compiled. If test scope doesn't work it seems like something is wrong.

jfdenise: @jamezp you were right, I fixed the pom.

jamezp (approved):

PR #374, opened 2023-09-05
[5.x] Add support for WildFly Glow

jfdenise:


plugin/src/main/java/org/wildfly/plugin/common/Utils.java
```diff
@@ -101,4 +118,66 @@ public static List<String> splitArguments(final CharSequence arguments) {
...
+            results.outputInformation(writer);
+        } catch (Exception ex) {
+            throw new MojoExecutionException(ex.getLocalizedMessage(), ex);
+        }
+        if (!dryRun) {
+            results.outputConfig(glowOutputFolder, false);
+        }
+        if (results.getErrorSession().hasErrors()) {
+            if (discoverProvisioningInfo.isFailsOnError()) {
+                throw new MojoExecutionException("Error detected by glow. Aborting.");
```
jamezp, line 175: Will errors be printed somewhere? What I mean by this, is will glow log some kind of error?

plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
```diff
@@ -195,6 +202,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             getLog().debug(String.format("Skipping " + getGoal() + " of %s:%s", project.getGroupId(), project.getArtifactId()));
             return;
         }
+        if (dryRun) {
+            getLog().info("Dry run execution, no server provisioned.");
```
jamezp, line 206: Minor, but I think we should say something like "Dry run execution, no server will be provisioned"

plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
```diff
@@ -254,30 +266,15 @@ private void provisionServer(Path home) throws ProvisioningException,
...
-                }
-                if (layers.isEmpty() && !STANDALONE_XML.equals(layersConfigurationFileName)) {
-                    throw new MojoExecutionException(
-                            "layers-configuration-file-name has been set although no layers are defined.");
+            ProvisioningConfig config = buildGalleonConfig(pm);
+            if (dryRun) {
+                Path targetPath = Paths.get(project.getBuild().getDirectory());
+                Path file = targetPath.resolve(PLUGIN_PROVISIONING_FILE);
+                getLog().info("Dry-run execution, generating provisioning.xml file: " + file);
+                try (FileWriter writer = new FileWriter(file.toFile())) {
```
jamezp, line 274: IMO we should use something like:
```java
try (BufferedWritter writer = Files.newBufferedWriter(file, StandardCharsets.UTF_8) {
```

plugin/src/main/java/org/wildfly/plugin/provision/GlowConfig.java
```diff
@@ -0,0 +1,177 @@
...
+    public GlowConfig() {
+    }
+
+    public Arguments toArguments(Path deployment, Path inProvisioning) {
+        Set<String> profiles = new HashSet<>();
+        if (profile != null) {
+            profiles.add(profile);
+        }
+        List<Path> lst = new ArrayList<>();
+        lst.add(deployment);
```
jamezp, line 54: If this is the only entry, we should just use `List.of(deployment);`.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -174,14 +177,81 @@ public class PackageServerMojo extends AbstractProvisionServerMojo {
     @Parameter(defaultValue = "false", property = PropertyNames.SKIP_PACKAGE_DEPLOYMENT)
     protected boolean skipDeployment;
 
+    /**
+     * Galleon provisioning info discovery. By enabling this feature, the set of Galleon feature-packs
+     * and layers are automatically discovered by scanning the deployed application.
+     * You can configure the following items:
```
jamezp, line 183: Minor again, but we should probably surround this in a `<p></p>`. It will just format better for the documentation.

jamezp (commented):
Just a couple minor changes/comments.

plugin/src/main/java/org/wildfly/plugin/common/Utils.java
jfdenise: Yes, glow has a pluggable message writer tha , in the plugin cases route logs to maven logs: https://github.com/jfdenise/wildfly-maven-plugin/blob/glow-only/plugin/src/main/java/org/wildfly/plugin/common/GlowMavenMessageWriter.java
BTW I am changing glow by "WildFly Glow"

plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
jfdenise: Ok, thanks.

plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
jfdenise: Ok, thanks.

plugin/src/main/java/org/wildfly/plugin/provision/GlowConfig.java
jfdenise: Ok, thanks.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
jfdenise: Ok (I hope what I did is what you suggested).

jfdenise:
Failure is unrelated to this change. Known instability when producing the docker image.

plugin/src/main/java/org/wildfly/plugin/dev/DevMojo.java
```diff
@@ -625,18 +743,32 @@ private void watch(final WatchService watcher, final DeploymentManager deploymen
...
                         }
+                        boolean repackaged = false;
                         if (remote || result.requiresRepackage()) {
+                            // The first packaging was not an exploded war, clean it.
+                            if (requiresWarDeletion) {
+                                final Path path = resolveWarLocation();
+                                deleteRecursively(path);
+                                requiresWarDeletion = false;
+                                repackaged = true;
+                            }
```
jamezp, line 754: We should likely perform an undeploy, at least if it's not remote, before we delete the path.

jamezp (commented):
Just one more minor thing. I would think we need to undeploy, for exploded content, before we delete the content.

jamezp:
I've also bumped main to 5.0.0.Alpha1-SNAPSHOT

plugin/src/main/java/org/wildfly/plugin/dev/DevMojo.java
jfdenise: @jamezp , thank-you. I found some issues with the remote goal that I fixed. When remote is enabled, this logic shouldn't be applied and  the reprovisioning/re-start shouldn't occur. I also found an issue with the directory name used by the war plugin when the "war" goal is used. In this case, the directory can't contain the .war extension. This extension has been added for Glow to understand what kind of exploded deployment we are dealing with. In the remote case, Glow scanning is meaningless.

jfdenise:
Failing test is un-related to this change.

jamezp (approved):

PR #377, opened 2023-09-11
Enforce overwriteProvisionedServer when layers discovery is enabled

jfdenise:


jmesnil (approved):
+1 this ensures that the `wildfly:dev` goal will work out of the box when Glow is enabled

jmesnil:
@jfdenise It might be worth adding a comment to https://github.com/wildfly/wildfly-maven-plugin/blob/9352aeb93c5750f8309851fe5491f09b13fc93e7/plugin/src/main/java/org/wildfly/plugin/dev/DevMojo.java#L319 about this. wdyt?

jfdenise:
@jmesnil , thanks, commen tadded.

jamezp (approved):

PR #397, opened 2023-10-25
[4.1] Update prospero and wildfly-channels versions

jfdenise:
Upstream PR #398


jfdenise:
@jamezp , we should wait for a possible other release of prospero to come soon. Will ping here.

jfdenise:
@jamezp , we are good with the versions we wanted to see updated.

PR #441, opened 2023-12-13
[WFMP-232] Add support for bootable JAR

jfdenise:


core/src/test/java/org/wildfly/plugin/core/Environment.java
```diff
@@ -16,7 +18,7 @@
  * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
  */
 @SuppressWarnings({ "WeakerAccess", "Duplicates" })
-class Environment {
+public class Environment {
```
jamezp, line 21: Is there a reason this needs to be public?

jamezp (approved):
I think this looks good. I'm going to file a JIRA and probably file a follow up PR with some refactoring to get ready for the split so we can test it before we do the split.

PR #461, opened 2023-12-22
Fix for WFMP-235, Introduce a start-jar goal

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-235

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/server/AbstractServerStartMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/server/AbstractServerStartMojo.java
@@ -10,60 +10,25 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicBoolean;
 
-import org.apache.maven.execution.MavenSession;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugin.logging.Log;
-import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.Parameter;
-import org.apache.maven.project.MavenProject;
-import org.eclipse.aether.RepositorySystem;
-import org.eclipse.aether.RepositorySystemSession;
-import org.eclipse.aether.repository.RemoteRepository;
-import org.jboss.as.controller.client.ModelControllerClient;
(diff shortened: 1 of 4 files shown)


jamezp:
I'm on PTO until the 2nd, but I want to think about this. I'll give it a review once I'm back if I don't get bored and review it sooner.  :)

plugin/src/main/java/org/wildfly/plugin/server/AbstractServerStartMojo.java
```diff
@@ -113,190 +78,27 @@ public abstract class AbstractServerStartMojo extends AbstractServerConnection {
...
-    protected MavenRepoManager createMavenRepoManager() throws MojoExecutionException {
-        return new MavenArtifactRepositoryManager(repoSystem, session, repositories);
-    }
-
-    protected ServerContext startServer(final ServerType serverType) throws MojoExecutionException, MojoFailureException {
-        final Log log = getLog();
-        init();
-
+    @Override
+    protected Path getServer() throws MojoExecutionException, MojoFailureException {
```
jamezp, line 95: I think we should call this `getServerHome()` or `getJBossHome()`.

plugin/src/main/java/org/wildfly/plugin/server/StartBootableJarMojo.java
```diff
@@ -0,0 +1,143 @@
...
+import org.wildfly.plugin.provision.PackageServerMojo;
+import org.wildfly.plugin.tools.bootablejar.BootableJarSupport;
+
+/**
+ * Starts a WildFly Application Server packaged as Bootable JAR.
+ * <p/>
+ * The purpose of this goal is to start a WildFly Application Server packaged as a Bootable JAR for testing during the maven
+ * lifecycle.
+ */
+@Mojo(name = "start-bootable", requiresDependencyResolution = ResolutionScope.RUNTIME)
```
jamezp, line 34: I'm just thinking a bit out loud here. Should the goal name be `start-jar`? Just thinking of something short to type is all, but I'm not sure I like `start-jar` :)

plugin/src/main/java/org/wildfly/plugin/server/AbstractServerStartMojo.java
jfdenise: Done thanks.

plugin/src/main/java/org/wildfly/plugin/server/StartBootableJarMojo.java
jfdenise: I much prefer start-jar, start-bootable is also a bit ambiguous, all server boot... So changed to start-jar.

jamezp (approved):

PR #527, opened 2024-05-03
Add an example of provisioning a specific version.

starksm64:
Fix the formatting of the channels in the provision mojo.


jamezp:
Sorry for the long delay on this. This will be resolved when 5.0.1.Final is released with #538.

PR #532, opened 2024-05-16
Fix for WFMP-255, Fix Galleon Options code snippet

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-255


jmesnil (approved):
thanks

PR #534, opened 2024-05-23
Fix for WFMP-257 and WFMP-258

jfdenise:
Issues:
* https://issues.redhat.com/browse/WFMP-257
* https://issues.redhat.com/browse/WFMP-258


plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
jamezp, line 1: None of the changes in this file should be needed.

plugin/src/main/java/org/wildfly/plugin/common/GlowMavenMessageWriter.java
```diff
@@ -15,7 +15,7 @@ public class GlowMavenMessageWriter implements GlowMessageWriter {
 
     private final Log log;
 
-    GlowMavenMessageWriter(Log log) {
+    public GlowMavenMessageWriter(Log log) {
```
jamezp, line 18: This change does not appear to be needed.

plugin/src/main/java/org/wildfly/plugin/provision/AbstractPackageServerMojo.java
```diff
@@ -0,0 +1,375 @@
...
+    protected boolean skip;
+
+    /**
+     * Skip deploying the deployment after the server is provisioned ({@code false} by default).
+     */
+    @Parameter(defaultValue = "false", property = PropertyNames.SKIP_PACKAGE_DEPLOYMENT)
+    protected boolean skipDeployment;
+
+    @Inject
+    protected OfflineCommandExecutor commandExecutor;
```
jamezp, line 167: This can be private.

jamezp (changes requested):
Could I get a little more information on why this change is needed? The JIRA's don't really say much :)

jfdenise:
Closing this PR, will open a PR for WFMP-257 only, WFMP-258 will be not done.

PR #531, opened 2024-05-15
Fix for WFMP-254, Dependency on maven-core should be provided

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-254

PR #536, opened 2024-05-31
Revert "Fix for WFMP-254, Dependency on maven-core should be provided"

jamezp:
Reverts wildfly/wildfly-maven-plugin#531


jamezp:
This will not be needed if #537 works, which it seems to locally.

PR #543, opened 2024-06-06
[WFMP-262]: Allow to add labels to the images.

ehsavoie:
* Adding a labels parameter.

Issue: https://issues.redhat.com/browse/WFMP-262

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
@@ -6,6 +6,7 @@ package org.wildfly.plugin.provision;
 
 import static java.lang.String.format;
 import static java.lang.String.join;
+import static org.wildfly.plugin.common.PropertyNames.WILDFLY_IMAGE_LABELS;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
```diff
@@ -95,6 +97,9 @@ public class ApplicationImageMojo extends PackageServerMojo {
     @Parameter(alias = "image")
     private ApplicationImageInfo image;
 
+    @Parameter(property = WILDFLY_IMAGE_LABELS, alias = "labels", required = false)
+    private List<String> labels = Collections.emptyList();
```
jamezp, line 101: Should this be a `Map<String, String>` instead so it can be defined like:
```xml
<labels>
    <version>1.0</version>
    <description>&quot;Test description &quot;</description>
</labels>
```

Also the alias and required aren't really needed.

plugin/src/main/java/org/wildfly/plugin/provision/ApplicationImageMojo.java
```diff
@@ -95,6 +97,9 @@ public class ApplicationImageMojo extends PackageServerMojo {
     @Parameter(alias = "image")
     private ApplicationImageInfo image;
 
+    @Parameter(property = WILDFLY_IMAGE_LABELS, alias = "labels", required = false)
```
jamezp, line 100: This needs some Java doc for the plugin documentation. In that it needs a `@since 5.0.2`.

jamezp:
I created #545 to get this in. Thank you @ehsavoie!

PR #546, opened 2024-06-25
Replace the docker/podman binary by Jib to build the image

jmesnil:


jamezp:
I like this idea. Thank you @jmesnil!

jamezp:
Do you feel we should move forward with this @jmesnil? I personally like the idea.

jmesnil:
A key improvement to make would be to build multi arch image (which Jib should be able to do based on https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-do-i-specify-a-platform-in-the-manifest-list-or-oci-index-of-a-base-image)

jmesnil:
The downside of that approach is that users can no longer add custom Docker `RUN` instructions to the application images and they no longer have a `Dockerfile` template to copy...
For this use case, we might want to keep support for `docker` or `podman` (flagged as Deprecated in the draft) and add the ability to point to a `Dockerfile` (which would replace the one generated by the Maven plug-in)

PR #554, opened 2024-08-26
Fix for WFMP-257, Glow support ignore channels

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-257


plugin/src/main/java/org/wildfly/plugin/provision/ChannelMavenArtifactRepositoryManager.java
```diff
@@ -89,6 +97,10 @@ public ChannelMavenArtifactRepositoryManager(List<ChannelConfiguration> channels
         this.system = system;
     }
 
+    public ChannelSession getChannelSession() {
```
jamezp, line 100: Does this need to be public?

tests/standalone-tests/pom.xml
```diff
@@ -82,7 +82,10 @@
             <artifactId>parsson</artifactId>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.wildfly.channel</groupId>
+            <artifactId>channel-core</artifactId>
```
jamezp, line 87: This looks like it should have `<scope>test</scope>`.

plugin/src/main/java/org/wildfly/plugin/provision/ChannelMavenArtifactRepositoryManager.java
jfdenise: This was a left over, I removed the method.

tests/standalone-tests/pom.xml
jfdenise: Good, fixed.

jamezp:
I'm not sure I follow what this means by "Ignore channels". Could you explain that a bit please @jfdenise?

jfdenise:
@jamezp , sure. In the Glow upgrade we got: https://github.com/wildfly/wildfly-glow/commit/3c2b34ed5ab5b152012bdac35db6238b06fdc053
Glow will use the configured channel to resolve the feature-packs. Without this upgrade, the hard coded versions of feature-packs known by Glow are used.
The changes in the ChannelMavenArtifactManager, are fixes found adding tests to cover channels with upstream feature-packs.

jfdenise:
@jamezp , yes, glow 1.1.1 contains fixes that allow for correct channel handling.

jamezp (approved):

PR #555, opened 2024-08-26
Fix for WFMP-266, Allow to configure glow discovery to not scan the deployment

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-266


jamezp (approved):

jamezp:
This looks good, I'm just debating whether it should go into 5.1 or 5.0.1. 

I'm wondering if we should somehow document what properties work with that version. I'm not sure how that looks yet though.

jfdenise:
@jamezp , I think that we should go to 5.1. I suspect possible more changes in the near future (eg: allow to deploy multiple deployments) that would push for such 5.1 release.

jfdenise:
@jamezp , I just see for now the multiple deployments one. It seems that it would deserve a wildfly proposal.

PR #561, opened 2024-09-09
Fix for WFMP-269, Support multiple deployments in the package goal

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-269

Updated documentation for multiple deployments and Glow (that was missing).

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/DeploymentResolution.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/DeploymentResolution.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.plugin.provision;
+
+import java.io.File;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.project.MavenProject;
+
+/**
+ * Resolve possible deployments from the Maven project dependencies.
+ */
+class DeploymentResolution {
+
+    private static final String WAR = "war";
+    private static final String EAR = "ear";
+    private static final String RAR = "rar";
+    private static final String JAR = "jar";
+    private static final Set<String> SUPPORTED_DEPLOYMENT_TYPES = Set.of(EAR, JAR, RAR, WAR);
+
+    static DeploymentResolution getInstance(MavenProject project) throws MojoExecutionException {
+        return new DeploymentResolution(project);
+    }
+
+    private final Map<String, File> files = new HashMap<>();
+
+    private DeploymentResolution(MavenProject project) throws MojoExecutionException {
+        for (Artifact artifact : project.getArtifacts()) {
+            if (SUPPORTED_DEPLOYMENT_TYPES.contains(artifact.getType())) {
+                files.put(buildKey(artifact.getGroupId(), artifact.getArtifactId(),
+                        artifact.getClassifier(), artifact.getType()), artifact.getFile());
+            }
+        }
+    }
+
+    File getFile(final String groupId, final String artifactId, final String classifier,
+            final String type) throws MojoExecutionException {
+        return files.get(buildKey(groupId, artifactId, classifier, type));
+    }
+
+    private String buildKey(String groupId, String artifactId, String classifier, String type) throws MojoExecutionException {
+        if (groupId == null) {
+            throw new MojoExecutionException("Deployment groupId can't be null");
+        }
+        if (artifactId == null) {
+            throw new MojoExecutionException("Deployment artifactId can't be null");
+        }
+        if (type == null) {
+            throw new MojoExecutionException("Deployment type can't be null");
+        }
+        if (!SUPPORTED_DEPLOYMENT_TYPES.contains(type)) {
+            throw new MojoExecutionException(
+                    "Deployment type " + type + " is not supported. Supported types are " + SUPPORTED_DEPLOYMENT_TYPES);
+        }
+        return groupId + ":" + artifactId + ":" + (classifier == null ? "" : classifier) + ":" + type;
+    }
+}
--- a/plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
@@ -17,7 +17,10 @@ import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 
 import javax.inject.Inject;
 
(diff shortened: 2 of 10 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -256,10 +260,34 @@ public class PackageServerMojo extends AbstractProvisionServerMojo {
...
+     *       &lt;artifactId&gt;deployment artifactId&lt;/artifactId&gt;
+     *       &lt;type&gt;ear|jar|rar|war&lt;/type&gt;
+     *       &lt;classifier&gt;Optional classifier&lt;/classifier&gt;
+     *       &lt;name&gt;Optional deployment name&lt;/name&gt;
+     *     &lt;/deployment&gt;
+     *   &lt;/deployments&gt;
+     * </pre>
+     */
+    @Parameter(alias = "deployments", required = false)
+    List<DeploymentConfiguration> deployments = new ArrayList<>();
```
jamezp, line 283: This should likely be private.

plugin/src/main/java/org/wildfly/plugin/provision/DeploymentResolution.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final String RAR = "rar";
+    private static final String JAR = "jar";
+    private static final Set<String> SUPPORTED_DEPLOYMENT_TYPES = Set.of(EAR, JAR, RAR, WAR);
+    private static final String TEST_JAR = "test-jar";
+
+    static DeploymentResolution getInstance(MavenProject project) throws MojoExecutionException {
+        return new DeploymentResolution(project);
+    }
+
+    private final Map<String, File> files = new TreeMap<>();
```
jamezp, line 32: What is the reason for a `TreeMap`?  It seems we could just use a `HashMap` which would perform a little better.

plugin/src/main/java/org/wildfly/plugin/provision/DeploymentResolution.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private DeploymentResolution(MavenProject project) throws MojoExecutionException {
+        for (Artifact artifact : project.getArtifacts()) {
+            if (TEST_JAR.equals(artifact.getType())) {
+                continue;
+            }
```
jamezp, line 38: I'm not sure this is needed. The `test-jar` type is not in the `SUPPORTED_DEPLOYMENT_TYPES` so ti doesn't seem like it should be needed.

plugin/src/main/java/org/wildfly/plugin/provision/DeploymentResolution.java
```diff
@@ -0,0 +1,67 @@
...
+            if (SUPPORTED_DEPLOYMENT_TYPES.contains(artifact.getType())) {
+                files.put(buildKey(artifact.getGroupId(), artifact.getArtifactId(),
+                        artifact.getClassifier(), artifact.getType()), artifact.getFile());
+            }
+        }
+    }
+
+    public File getFile(final String groupId, final String artifactId, final String classifier,
```
jamezp, line 46: This should be package-private since the type itself is.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -305,9 +342,45 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             getLog().debug(String.format("Skipping " + getGoal() + " of %s:%s", project.getGroupId(), project.getArtifactId()));
             return;
         }
+        deploymentResolution = DeploymentResolution.getInstance(project);
         super.execute();
     }
 
+    private void deploy(Path deploymentContent, String targetName) throws MojoDeploymentException {
+        if (Files.exists(deploymentContent)) {
```
jamezp, line 350: Shouldn't we be doing something like throwing an error if the file does not exist? Or at least logging a warning/error.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -305,9 +342,45 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+    private Map<String, Path> getDeployments() throws MavenUniverseException, MojoExecutionException {
+        if (deploymentPaths == null) {
+            deploymentPaths = new LinkedHashMap<>();
+            for (DeploymentConfiguration c : deployments) {
+                Path p;
+                File f = deploymentResolution.getFile(c.getGroupId(), c.getArtifactId(), c.getClassifier(),
+                        c.getType());
+                if (f == null) {
+                    throw new MojoExecutionException("Deployment not found " + c.getGroupId() + ":" + c.getArtifactId()
+                            + (c.getClassifier() == null ? "" : ":" + c.getClassifier()) + ":" + c.getType());
```
jamezp, line 374: Minor, but probably useful. We could do this logic in the `DeploymentConfiguration.toString()`.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -492,6 +564,9 @@ private void warnExtraConfig(Path extraContentDir) {
 
     protected Path getDeploymentContent() throws MojoExecutionException {
         final PackageType packageType = PackageType.resolve(project);
+        if (packageType.getPackaging().equals("pom")) {
```
jamezp, line 567: This should probably use `packageType.isIgnored()`.

jamezp (changes requested):

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
jfdenise: Fixed

plugin/src/main/java/org/wildfly/plugin/provision/DeploymentResolution.java
jfdenise: Good catch.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
jfdenise: Good catch, logging a warning.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
jfdenise: Yes!

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
jfdenise: Done.

plugin/src/main/java/org/wildfly/plugin/provision/DeploymentResolution.java
jfdenise: A letfover... Good catch.

jfdenise:
@jamezp I implemented your great comments. Thank-you!

jfdenise:
@torsten-liermann , it would be great if you could validate that this solution works for you. You have been the main source of requirement for it!
Thank-you.

torsten-liermann:
Thank you for the quick implementation. The implementation works as expected for the example project. I would have designed the idea similarly to the maven-ear-plugin: deploying all dependencies defined in the pom.xml and structuring the "deployments" configuration analogous to the "modules" configuration.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
jamezp: This doesn't appear to be changed.

jamezp (changes requested):

jfdenise:
@jamezp , done. Thank-you.

jamezp:
This was replaced by #604

PR #567, opened 2024-09-10
Fix for WFMP-270, Plugin description should specify how to build a Hollow JAR

jfdenise:
Issue: https://issues.redhat.com/browse/WFMP-270

In addition, added documentation on how to produce a bootable JAR.

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
@@ -227,7 +227,8 @@ public class PackageServerMojo extends AbstractProvisionServerMojo {
     private GlowConfig discoverProvisioningInfo;
 
     /**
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -227,7 +227,8 @@ public class PackageServerMojo extends AbstractProvisionServerMojo {
     private GlowConfig discoverProvisioningInfo;
 
     /**
-     * Package the provisioned server into a WildFly Bootable JAR.
+     * Package the provisioned server into a WildFly Bootable JAR. In order to produce an hollow jar (a jar that doesn't contain
```
jamezp, line 230: Minor, but it should be "a hollow jar".

jamezp (approved):

PR #569, opened 2024-09-16
[WFMP-275] Add GPG checks to the channel provisioning

spyrkob:
Issue: https://issues.redhat.com/browse/WFMP-275


jamezp:
@spyrkob Do we want this in 5.1.0.Final? If so, could you update this PR?

PR #604, opened 2024-11-27
[WFMP-269] Allow multiple deployments to be used for the package goal

jamezp:
This is a replacement for #561

@jfdenise WDYT? The idea is you don't need include a deployment definition for each extra deployment. For example in the demo project, you'd simple add:

```xml
<included-dependencies>
    <include>::ear</include>
</included-dependencies>
```

**OR**

```xml
<included-dependencies>
    <include>demo.multiapps</include>
</included-dependencies>
```


jamezp:
Maybe it should be `included-deployments` instead of `included-dependencies`.

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
```diff
@@ -271,6 +279,22 @@ public class PackageServerMojo extends AbstractProvisionServerMojo {
     @Parameter(alias = "bootable-jar-install-artifact-classifier", property = PropertyNames.BOOTABLE_JAR_INSTALL_CLASSIFIER, defaultValue = BootableJarSupport.BOOTABLE_SUFFIX)
     private String bootableJarInstallArtifactClassifier;
 
+    /**
+     * A list of the dependencies to include as deployments. The dependencies must be defined as dependencies in the
```
jfdenise, line 283: We should perhaps add the syntax of the dependency?

jfdenise:
@jamezp , I like the dependencies in the name, this is what they are, no confusion with File path,...
perhaps deployed-dependencies?

plugin/src/main/java/org/wildfly/plugin/provision/PackageServerMojo.java
jamezp: Good call. I meant to add that and completely forgot.

jfdenise (approved):

torsten-liermann:
I liked the approach with the dependencies because it corresponds, for example, to the maven-war and maven-ear plugins. In my case, what needs to be deployed comes from the Maven build and can always be easily referenced as a Maven artifact. From my understanding as a long-time Maven user, this is simply consistent.

jamezp:
@torsten-liermann I'm not sure I follow. This approach does use dependencies. You've just to got explicitly state what you want included.

PR #609, opened 2024-12-03
[WFMP-283] Use a version range for the WildFly Channel utilities.

jamezp:
https://issues.redhat.com/browse/WFMP-283

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@
         <version.org.wildfly.core>25.0.2.Final</version.org.wildfly.core>
         <version.org.wildfly.launcher>1.0.0.Beta3</version.org.wildfly.launcher>
         <version.org.wildfly>32.0.1.Final</version.org.wildfly>
-        <version.org.wildfly.channel>1.2.1.Final</version.org.wildfly.channel>
+        <version.org.wildfly.channel>[1.2.1.Final,)</version.org.wildfly.channel>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -80,7 +80,7 @@
         <version.org.wildfly.core>25.0.2.Final</version.org.wildfly.core>
         <version.org.wildfly.launcher>1.0.0.Beta3</version.org.wildfly.launcher>
         <version.org.wildfly>32.0.1.Final</version.org.wildfly>
-        <version.org.wildfly.channel>1.2.1.Final</version.org.wildfly.channel>
+        <version.org.wildfly.channel>[1.2.1.Final,)</version.org.wildfly.channel>
```
jfdenise, line 83: I have a mixed feeling, we could end up with an invalid plugin, other dependencies depending on the channel could be impacted. We can't really test the plugin at build time. I would prefer to have a fixed version.

We should constrain to the next major? We could have API change that break the plugin, could be something like [1.2.1.Final,2.0.0.Final).

The resolution is also bound to the set of maven repositories configured. In case we have redhat GA repository, the -redhat-XXXX versions could be selected. The range should filter them out.

jamezp: Yeah. I don't really like this either. However, we do need some kind of solution as an upgrade is required for any projects attempting to test against the latest WildFly. I'm just trying to think of a way to avoid this in the future.

PR #612, opened 2024-12-03
Upgrade prospero-metadata to 1.3.0.Beta2

TomasHofman:
This is the prospero-metadata version minor relevant to next EAP stream.

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/ChannelMavenArtifactRepositoryManager.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/ChannelMavenArtifactRepositoryManager.java
@@ -186,8 +186,7 @@ public class ChannelMavenArtifactRepositoryManager implements MavenRepoManager, 
 
     public void done(Path home) throws MavenUniverseException, IOException {
         ChannelManifest channelManifest = channelSession.getRecordedChannel();
-        final ManifestVersionRecord currentVersions = new ManifestVersionResolver(localCachePath, system)
-                .getCurrentVersions(channels);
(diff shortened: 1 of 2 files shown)


TomasHofman:
This needs to be reviewed by @spyrkob .

spyrkob (approved):

plugin/src/main/java/org/wildfly/plugin/provision/ChannelMavenArtifactRepositoryManager.java
```diff
@@ -186,8 +186,7 @@ private void resolveFromChannels(MavenArtifact artifact) throws UnresolvedMavenA
 
     public void done(Path home) throws MavenUniverseException, IOException {
         ChannelManifest channelManifest = channelSession.getRecordedChannel();
-        final ManifestVersionRecord currentVersions = new ManifestVersionResolver(localCachePath, system)
-                .getCurrentVersions(channels);
+        final ManifestVersionRecord currentVersions = ManifestVersionResolver.getCurrentVersions(channelSession);
```
jamezp, line 189: Is there a reason for this change?

spyrkob: @jamezp I deprecated the old method - it was relying on a hack I want to get rid of. The new method uses changed wildfly-channel API to achieve the same goal more cleanly.

TomasHofman: Yeah, that was really @spyrkob 's request.

PR #620, opened 2024-12-06
Add commons-logging-jboss-logmanager dependency to test to avoid CNF Exception

jfdenise:
The Exception doesn't make the tests to fail but is displayed in the test logs.


jamezp:
This is because the `<java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>` is set in the surefire configuration. I've removed that in #621 

The other option is to add dependency to the `org.jboss.logmanager:jboss-logmanager` and a configuration file. Just removing the property seems fine for now though.

The reason this works is because of a transitive dependency on the log manager. However, we'd want a direct dependency and not use the commons-logging.

PR #640, opened 2025-01-15
Fix for WFMP-290, Upgrade to Glow 1.3.0.Final

jfdenise:
This upgrade makes WildFly Glow to retrieve its metadata from maven, that is something that blocks some users to use the WildFly Maven Plugin with Glow feature (in some pipeline context where only maven repositories are allowed to connect to).


jamezp (approved):

jamezp:
Note typically we would not do a minor upgrade in a patch release. However, this appears to be compatible.

PR #683, opened 2025-07-16
[WFMP-304][JBEAP-28721] Bootable JAR won't be created if server provisioning-dir already exists

lvydra:
https://issues.redhat.com/browse/WFMP-304
https://issues.redhat.com/browse/JBEAP-28721


jamezp:
I'm not sure this is correct. We shouldn't expose the `bootableJar` parameter outside of the `package` goal. It looks like all this does is log another message.

What is the desired result for an issue like this? I'm not entirely following the error.

lvydra:
Hi @jamezp, thank you for looking at it. 
I'm not sure either. 

> What is the desired result for an issue like this? I'm not entirely following the error.

There are some cases when bootable jar creation is skipped due to an already existing provisioning-dir. The purpose of this issue is to notify the user about that. Look at the discussion on https://issues.redhat.com/browse/JBEAP-28721
If you think this is not necessary, feel free to close this PR and the issues as well.

jfdenise:
Making change in the target directory can lead to un-expected results. A user should call clean.
I think that we could fix the issue. If a bootable jar doesn't exist although the provisioning dir exists, we should just produce the bootable JAR. I would not add the bootableJar field to the provisioning goal. That is really the package goal that defines it. So we would need to refactor a bit the goals to detect the case and build a bootable JAR based on the existing directory.

PR #770, opened 2026-04-13
Fix for WFMP-345, Upgrade wildfly common to 2.0.1

jfdenise:
Issue: https://redhat.atlassian.net/browse/WFMP-345


jamezp:
Removed the dependency in #771. Thank you @jfdenise

PR #774, opened 2026-04-14
Fix for WFMP-348, Add support for variant in WildFly Glow configuration

jfdenise:
Issue: https://redhat.atlassian.net/browse/WFMP-348

Incorporates a bump to WildFly Glow 2.0.0.Beta2

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/GlowConfig.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/GlowConfig.java
@@ -9,6 +9,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.logging.Log;
 import org.wildfly.glow.Arguments;
 import org.wildfly.glow.OutputFormat;
 import org.wildfly.glow.ScanArguments.Builder;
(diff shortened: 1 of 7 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/GlowConfig.java
```diff
@@ -29,20 +30,28 @@ public class GlowConfig {
...
     public GlowConfig() {
     }
 
-    public Arguments toArguments(List<Path> lst, Path inProvisioning, String layersConfigurationFileName) {
+    public Arguments toArguments(Log log, List<Path> lst, Path inProvisioning, String layersConfigurationFileName)
+            throws Exception {
         final Set<String> profiles = profile != null ? Set.of(profile) : Set.of();
         List<Path> deployments = ignoreDeployment ? Collections.emptyList() : lst;
+        if (serverVariant != null && preview) {
+            throw new Exception("Preview can't be set when a server variant has been set.");
```
jamezp:
```suggestion
            throw new MojoExecutionException("Preview can't be set when a server variant has been set.");
```

jamezp: The other option would be an `IllegalArgumentException`, but a generic `Exception` doesn't seem right.

tests/standalone-tests/src/test/java/org/wildfly/plugin/provision/PackageTest.java
```diff
@@ -124,6 +124,27 @@ public void testGlowNoDeploymentPackage(final Mojo packageMojo) throws Exception
...
+        Path jbossHome = resolvePath("packaged-preview-glow-server");
+        TestEnvironment.checkStandaloneWildFlyHome(jbossHome, 1, layers, null, true);
+    }
+
+    @Test
+    @InjectMojo(goal = "package", pom = "package-unknown-variant-glow-pom.xml")
+    public void testGlowUnknownVariantPackage(final Mojo packageMojo) throws Exception {
+        try {
+            packageMojo.execute();
+            throw new Exception("Execution should have failed");
```
jamezp:
```suggestion
            Assertions.fail("Execution should have failed");
```

jamezp (changes requested):

plugin/src/main/java/org/wildfly/plugin/provision/GlowConfig.java
jfdenise: Thank-you, resolved.

tests/standalone-tests/src/test/java/org/wildfly/plugin/provision/PackageTest.java
jfdenise: Thank-you, resolved.

jamezp (approved):

PR #780, opened 2026-05-09
[WFMP-353] Provision from file path directly to preserve <feature> elements

torsten-liermann:
Issue: https://redhat.atlassian.net/browse/WFMP-353

## Summary

When the `provision` goal is configured with a `provisioning-file` and no
plugin-side `<feature-packs>`, inline `<feature>` elements nested under
`<config>` are silently dropped. The plugin currently parses the file
into a `GalleonProvisioningConfig`, then calls
`Provisioning.provision(GalleonProvisioningConfig, ...)`. Inside the
isolated galleon-core class loader, the
`GalleonProvisioningConfig` → `ProvisioningConfig` converter only copies
layers, properties and config-deps from `GalleonConfigurationWithLayers`
and discards the `ConfigItemContainer` items (features, feature-groups)
carried by the parsed `ConfigModel`.

User-visible symptom: a `provisioning.xml` like

```xml
<config name="standalone.xml" model="standalone">
    <layers><include name="cloud-server"/></layers>
    <feature spec="subsystem.undertow.configuration.filter.expression-filter">
        <param name="expression-filter" value="my-filter"/>
        <param name="expression" value="path('/x') -> rewrite('/y')"/>
    </feature>
</config>
```

provisions a server in which the filter is missing, with no warning in
the build log. Even an intentionally invalid spec name does not fail the
build.

This patch routes the provisioning file directly to
`Provisioning.provision(Path, options)` when no plugin-side feature-packs
are configured. The internal `ProvisioningConfig` is fed into
`ProvisioningManager` without crossing the API boundary, so `<feature>`,
`<feature-group>`, `<packages>` and `<config-deps>` are preserved. The
dry-run path copies `provisioning.xml` verbatim to the target directory
instead of round-tripping through `storeProvisioningConfig`.

## Verification

1. Create a `provisioning.xml` with the snippet above.
2. Bind the `provision` goal in the `package` phase, point
   `provisioning-file` at the file, leave `<feature-packs>` empty.
3. Build with the unpatched plugin: `BUILD SUCCESS`, but
   `standalone/configuration/standalone.xml` contains no
   `expression-filter` and `target/server/.galleon/provisioning.xml`
   records only `<layers>`, no `<feature>`.
4. Build with this patch: `BUILD SUCCESS`, the filter shows up in the
   generated `standalone.xml`; an invalid spec name now fails the build
   with `Failed to locate feature spec ...`.

## Risk

Limited to the `provisioning-file` + empty `<feature-packs>` configuration
path. Other provisioning paths (channel-based, plugin-configured
feature-packs) are unchanged. The dry-run fallback is a verbatim file
copy, which is closer to the user's input than the previous
`storeProvisioningConfig` round-trip.

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
@@ -13,6 +13,7 @@ import java.net.MalformedURLException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
```diff
@@ -305,8 +306,22 @@ private void provisionServer(Path home) throws ProvisioningException,
...
+        // (loadProvisioningConfig -> provision(GalleonProvisioningConfig))
+        // silently drops <feature> elements nested under <config> because the
+        // GalleonProvisioningConfig <-> ProvisioningConfig conversion erases
+        // ConfigItemContainer items when the configuration crosses the
+        // isolated core class loader. Using the path-based provision call
+        // preserves the parsed ProvisioningConfig as-is.
+        Path resolvedProvisioningFile = resolvePath(project, provisioningFile.toPath());
+        boolean provisionFromFile = featurePacks.isEmpty()
+                && Files.exists(resolvedProvisioningFile);
         GalleonProvisioningConfig config = buildGalleonConfig(galleonBuilder);
```
jfdenise, line 321: Good catch! I feel that the handling of the provisioning.xml file there https://github.com/wildfly/wildfly-maven-plugin/blob/main/plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java#L345
becomes useless. Perhaps a bit of refactoring to avoid this useless code.

jfdenise (approved):
Except from the possible refactoring I mentioned that could be handled in another PR, I am approving this PR.

plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
jamezp: +1 to refactoring, but I'm okay with a follow up PR.

jamezp (approved):

PR #783, opened 2026-05-12
fix: inject session proxy into Aether repositories before embedded resolves

torsten-liermann:
# fix: inject session proxy into Aether repositories before embedded resolves

## Problem
Aether's `HttpTransporter` reads `repository.getProxy()` directly; it does not consult `session.getProxySelector()`. The Maven CLI normally runs `RepositorySystem.newResolutionRepositories(session, repos)` before Mojo invocation, which copies the session-level proxy onto every `RemoteRepository`. Embedded callers (`maven-plugin-testing-harness` 3.5.1, custom invokers) skip that step, so `repository.getProxy() == null` and downstream resolves connect directly.

## Fix
New helper `MavenRepositoriesEnricher.injectSessionProxies(repoSystem, session, repos)` wraps the `newResolutionRepositories` call. Invoked next to each `MavenRepositoriesEnricher.enrich` call site that the plugin uses before building its own `MavenArtifactRepositoryManager`:

- `AbstractProvisionServerMojo.enrichRepositories` — provision / package goals
- `ExecuteCommandsMojo.execute` — `execute-commands` goal (`LocalCLIExecutor.resolveCLI` resolves `wildfly-cli` on its own through the resulting `MavenArtifactRepositoryManager`)
- `AbstractStartMojo.init` — covers every `AbstractStartMojo` subclass (`StartMojo`, `RunMojo`, `DevMojo`, `StartJarMojo`)

Idempotent for the Maven CLI.

## Reproducer
`maven-plugin-testing-harness` 3.5.1 + Maven Resolver 1.9.27 + fresh local m2. Run any of the goals above behind an HTTP proxy declared in `~/.m2/settings.xml`:

```
Could not transfer artifact ... : Connect timed out
```

`session.getProxySelector().getProxy(repo)` returns the configured proxy, but `HttpTransporter` never sees it because `repository.getProxy() == null`.

### DIFF
--- a/core/src/main/java/org/wildfly/plugin/core/MavenRepositoriesEnricher.java
+++ b/core/src/main/java/org/wildfly/plugin/core/MavenRepositoriesEnricher.java
@@ -105,7 +105,8 @@ public class MavenRepositoriesEnricher {
     }
 
     /**
-     * Inject the session-level proxy/mirror/auth onto each repository.
+     * Returns the repository list to use for resolution, with the session-level
+     * proxy/mirror/auth injected onto every entry.
      *
      * Aether's HttpTransporter reads {@code repository.getProxy()} directly
      * and never consults {@code session.getProxySelector()}. The Maven CLI
@@ -116,18 +117,21 @@ public class MavenRepositoriesEnricher {
      * Call this helper after {@link #enrich} so downstream resolves honour
      * the settings.xml proxy regardless of the calling context.
      *
-     * @param repoSystem   the Aether repository system (may be null - call is a no-op)
-     * @param session      the Aether repository session (may be null - call is a no-op)
-     * @param repositories the repository list to enrich in place
+     * The input list is treated as read-only and never mutated. When injection
+     * cannot or need not be performed (null inputs, empty list), the input
+     * list itself is returned unchanged.
+     *
+     * @param repoSystem   the Aether repository system (may be {@code null} - input list is returned unchanged)
+     * @param session      the Aether repository session (may be {@code null} - input list is returned unchanged)
+     * @param repositories the repository list to enrich; never mutated, may be {@code null} or empty
+     * @return the enriched repository list when injection is possible; otherwise the original input list
      */
-    public static void injectSessionProxies(RepositorySystem repoSystem, RepositorySystemSession session,
-            List<RemoteRepository> repositories) {
+    public static List<RemoteRepository> injectSessionProxies(RepositorySystem repoSystem,
+            RepositorySystemSession session, List<RemoteRepository> repositories) {
         if (repoSystem == null || session == null || repositories == null || repositories.isEmpty()) {
-            return;
+            return repositories;
         }
-        final List<RemoteRepository> enriched = repoSystem.newResolutionRepositories(session, repositories);
-        repositories.clear();
-        repositories.addAll(enriched);
+        return repoSystem.newResolutionRepositories(session, repositories);
     }
 
     private static Set<String> getUrls(List<RemoteRepository> repositories) {
--- a/core/src/main/java/org/wildfly/plugin/core/MavenRepositoriesEnricher.java
+++ b/core/src/main/java/org/wildfly/plugin/core/MavenRepositoriesEnricher.java
@@ -19,6 +19,8 @@ import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.settings.Proxy;
 import org.apache.maven.settings.Settings;
+import org.eclipse.aether.RepositorySystem;
+import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.repository.RemoteRepository;
 import org.eclipse.aether.repository.RepositoryPolicy;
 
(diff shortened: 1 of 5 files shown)


torsten-liermann:
Marking as draft — the fix is incomplete. `LocalCLIExecutor.resolveCLI` is a second resolution path that bypasses `AbstractProvisionServerMojo.enrichRepositories` and still hits the same proxy issue. Follow-up patch needed before this can land.

core/src/main/java/org/wildfly/plugin/core/MavenRepositoriesEnricher.java
```diff
@@ -102,6 +104,32 @@ public static void enrich(MavenSession session, MavenProject project, List<Remot
         }
     }
 
+    /**
+     * Inject the session-level proxy/mirror/auth onto each repository.
+     *
+     * Aether's HttpTransporter reads {@code repository.getProxy()} directly
```
jfdenise, line 111: @torsten-liermann , would you share the link to Aether's HttpTransporter source code?

torsten-liermann:
@jfdenise sure — `HttpTransporter` (Maven Resolver 1.9.27):

- Constructor reads the proxy off the repository directly at line 187:
  https://github.com/apache/maven-resolver/blob/maven-resolver-1.9.27/maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/HttpTransporter.java#L187
  ```java
  this.proxy = toHost(repository.getProxy());
  ```
- That `HttpHost` is then passed to Apache HttpClient at line 340:
  https://github.com/apache/maven-resolver/blob/maven-resolver-1.9.27/maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/HttpTransporter.java#L340

`session.getProxySelector()` is never consulted here. The Maven CLI applies it earlier, via `DefaultRepositorySystem.newResolutionRepositories`:
https://github.com/apache/maven-resolver/blob/maven-resolver-1.9.27/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystem.java
which rebuilds each `RemoteRepository` with `setProxy(session.getProxySelector().getProxy(repo))`. Embedded callers that skip that step end up with `repository.getProxy() == null`, which is the case this PR addresses.

core/src/main/java/org/wildfly/plugin/core/MavenRepositoriesEnricher.java
```diff
@@ -102,6 +104,32 @@ public static void enrich(MavenSession session, MavenProject project, List<Remot
...
+     * @param session      the Aether repository session (may be null - call is a no-op)
+     * @param repositories the repository list to enrich in place
+     */
+    public static void injectSessionProxies(RepositorySystem repoSystem, RepositorySystemSession session,
+            List<RemoteRepository> repositories) {
+        if (repoSystem == null || session == null || repositories == null || repositories.isEmpty()) {
+            return;
+        }
+        final List<RemoteRepository> enriched = repoSystem.newResolutionRepositories(session, repositories);
+        repositories.clear();
```
jfdenise: Clearing this list doesn't make me comfortable. Imagine the list is read only. I think that we should return a new list and possibly refactor the method. WFYT?

torsten-liermann: @jfdenise Agreed — junior's work without review

jfdenise: All your recent potential contributions are far from junior ones, really small things to fix.

plugin/src/main/java/org/wildfly/plugin/provision/AbstractProvisionServerMojo.java
```diff
@@ -293,8 +293,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         }
     }
 
-    protected void enrichRepositories() throws MojoExecutionException {
+    protected List<RemoteRepository> enrichRepositories() throws MojoExecutionException {
```
jfdenise, line 296: That is a breaking change , that is fine to introduce in new 6.x final. @jamezp WDYT?

jamezp: Oops, I missed this and just did a release today because we needed it for WildFly 40. If we feel we need this though we can always do a 6.1.

core/src/main/java/org/wildfly/plugin/core/MavenRepositoriesEnricher.java
```diff
@@ -102,6 +104,36 @@ public static void enrich(MavenSession session, MavenProject project, List<Remot
...
+     * The input list is treated as read-only and never mutated. When injection
+     * cannot or need not be performed (null inputs, empty list), the input
+     * list itself is returned unchanged.
+     *
+     * @param repoSystem   the Aether repository system (may be {@code null} - input list is returned unchanged)
+     * @param session      the Aether repository session (may be {@code null} - input list is returned unchanged)
+     * @param repositories the repository list to enrich; never mutated, may be {@code null} or empty
+     * @return the enriched repository list when injection is possible; otherwise the original input list
+     */
+    public static List<RemoteRepository> injectSessionProxies(RepositorySystem repoSystem,
```
slawekjaranowski, line 129: Instead of creating new method, method: `MavenRepositoriesEnricher#enrich` can be refactored to use only `repoSystem.newResolutionRepositories` and all manual works with settings proxy can be removed

PR #784, opened 2026-05-12
[WFMP-354] perf(provision): bulk-resolve Galleon artifacts via Aether

torsten-liermann:
Issue: https://redhat.atlassian.net/browse/WFMP-354

# perf(provision): bulk-resolve Galleon artifacts via Aether

## Change
`ChannelMavenArtifactRepositoryManager` overrides `MavenRepoManager.resolveAll(Collection)`. Concrete-version artifacts are pre-fetched in one `RepositorySystem.resolveArtifacts(...)` call; Aether's connector thread pool (default 5) downloads them in parallel. The per-artifact `resolve(MavenArtifact)` path is kept as fallback so channel-recording semantics are preserved and any artifact missed by the bulk pre-fetch is retried individually.

The bulk `resolveAll` path is gated by Galleon option `<jboss-bulk-resolve-artifacts>true</jboss-bulk-resolve-artifacts>`. Enabled in `tests/pom.xml` so it is exercised by the existing provision-server fixture. Without the option the per-artifact `resolve(MavenArtifact)` flow is unchanged.

Independent of that option, the repository session is now constructed as `new DefaultRepositorySystemSession(contextSession)` instead of `MavenRepositorySystemUtils.newSession() + setLocalRepositoryManager(...)`. This affects every resolution path through `ChannelMavenArtifactRepositoryManager`: it inherits the caller's mirror/proxy/auth selectors, not only its local repository manager.

## Cold-cache measurement

| Phase | Before | After |
|---|---|---|
| Galleon provisioning | 172.2 s | 77.3 s (-55%) |
| `mvn install` end-to-end | 218.6 s | 128.2 s (-41%) |

Warm-cache runs unchanged.

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/ChannelMavenArtifactRepositoryManager.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/ChannelMavenArtifactRepositoryManager.java
@@ -13,6 +13,7 @@ import java.nio.file.FileSystem;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/ChannelMavenArtifactRepositoryManager.java
```diff
@@ -103,6 +104,65 @@ public ChannelMavenArtifactRepositoryManager(List<ChannelConfiguration> channels
...
+                // Partial failure is fine: the per-artifact resolve below retries the misses
+                // through the channel/direct-resolve fallback.
+                if (log.isDebugEnabled()) {
+                    log.debug("Bulk pre-fetch reported missing artifacts; falling back to per-artifact resolve. "
+                            + ex.getMessage());
+                }
+            }
+        }
+        // Fall back to the per-artifact path; cached artifacts are now local hits.
+        for (MavenArtifact a : artifacts) {
```
jfdenise, line 161: Shouldn't we only do this fall back in case an error occured?

jamezp (approved):

jamezp:
Thank you @torsten-liermann

PR #788, opened 2026-05-18
[WFLY-307] Start with remote debugging enabled in dev mode by default.

jamezp:
https://redhat.atlassian.net/browse/WFMP-307


alf:
It seems like this change breaks `<debug>true</debug>`.

We've been using `<debug>` combined with `<debugPort>` to get a jdwp listening on port 5005. Now that crashes. I've been trying to affect the port this default debugger listens on but there doesn't really seem to be a way to do that.

I suggest rolling back this change and releasing 6.0.1.

jamezp:
I see the issue, sorry about that @alf . I've filed https://redhat.atlassian.net/browse/WFMP-361 and will get a release out as soon as it's fixed.

PR #790, opened 2026-05-21
doc: add link to Galleon guide

jmesnil:
to provide the list of available options for the galleon-options property


jamezp (approved):

jamezp:
Merging as we don't need CI for this since it's a JavaDoc only change.

Thank you @jmesnil

PR #809, opened 2026-07-22
Fix for WFMP-365, Bulk resolution with channel doesn't resolve artifact version from the channel

jfdenise:
ISSUE: https://redhat.atlassian.net/browse/WFMP-365

### DIFF
--- a/plugin/src/main/java/org/wildfly/plugin/provision/MavenArtifactMapper.java
+++ b/plugin/src/main/java/org/wildfly/plugin/provision/MavenArtifactMapper.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.plugin.provision;
+
+import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/wildfly/plugin/provision/MavenArtifactMapper.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import org.jboss.galleon.universe.maven.MavenUniverseException;
+import org.wildfly.channel.ArtifactCoordinate;
+import org.wildfly.channel.MavenArtifact;
+
+public class MavenArtifactMapper {
```
jamezp: Will this be used outside this package? If not, I'd suggest making it package private is all.

jfdenise: Thank-you @jamezp , fixed.

jamezp (approved):

